PR #2, opened 2017-08-07
Feature enter name

logicgraphy:
This commit includes basic implementation of redux to work with react within the application. This commit also includes functionality for a user to input/update their names (first, middle and last) and display it in the summary. 
The components with redux container features are moved to 'container' directory. Relevant Action and Reducer functions are implemented required by redux. Code organization will be an ongoing process, Action types will need to be maintained as a separate growing list, and Reducers will be clubbed together based on modules rather than individual feature.

Code review requested. @baccigalupi


baccigalupi:
Wondering why the codeclimate check failed? It looks like everything is good when I follow the link. Working on the PR now, but may be interrupted.

logicgraphy:
It was giving duplicate code issue because of input text field used three times for first, middle and last names respectively. Not sure why its not showing up anymore.

PR #3, opened 2017-08-08
[Finishes #149905850] Updates font for whole system

baccigalupi:
Added Source Sans Pro font as family font.


logicgraphy:
All test cases passed. Merging with the base.

PR #8, opened 2017-08-14
Features: Contact (Email address & Phone number)

logicgraphy:
Implements features for user to enter email address and phone number. Both features were part of same route, therefore merged into single pull request. 
[Completes: #149610077 and #149610303]


client/containers/contact.jsx
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+import React, {PropTypes} from 'react';
```
baccigalupi: I think we don't need PropTypes, or Link.

client/containers/contact.jsx
```diff
@@ -0,0 +1,40 @@
...
+const Contact = (props) => {
+  return (
+    <ContactForm onSubmit={props.onSubmit} onChange={props.onChange} contact={props.contact} />
+  );
+};
+
+function mapStateToProps(state) {
+  return {contact: state.contact};
+}
+
```
baccigalupi: In my pull request I wrapped up a lot of this logic, so that we aren't copying and pasting whole files. We don't have to do this in this commit, but we need to do a refactor after all the pieces get in.

client/containers/summary-handler.jsx
```diff
@@ -4,24 +4,40 @@ import React, {Component} from 'react';
...
       props.legalName.lastName !== '') {
-    return (<SummaryNames legalName={props.legalName} />);
+    summaryStack.push(<SummaryNames key="legalNameKey" legalName={props.legalName} />);
+  }
+  if(props.contact.emailAddress !== '' || props.contact.phoneNumber !== ''){
+    summaryStack.push(<SummaryContact key="contactKey" contact={props.contact}/>);
   }
   else{
-    return (<SummaryEmpty />);
+   summaryStack.push(<SummaryEmpty key="emptyKey"/>);
```
baccigalupi: I switched this to be another if statement where if the stack is empty, then add the empty one. Also, not sure the work 'Key' appended makes sense. Would be good to just have the purpose of the component:

test/features/email-address.feature
```diff
@@ -0,0 +1,34 @@
+Feature: I want to enter my email address
```
baccigalupi: Indentations are off in this file. Let's make sure that we are using 2 spaces instead of tabs and that the format is like the other files.

test/features/phone-number.feature
```diff
@@ -0,0 +1,34 @@
+Feature: I want to enter my phone number
+	As a DMV customer
```
baccigalupi: Here too, check the indentation. It looks crazy on github but not sure about the original files.

test/features/step_definitions/enter-email-address-steps.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+const assert = require('assert');
+
+module.exports = function(world) {
+  let browser = world.browser;
+
+  // world.given('I go to the new online DL application page', function(done) {
```
baccigalupi: Let's get rid of all the commented out code

test/features/step_definitions/enter-phone-number-steps.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+
+const assert = require('assert');
+
+module.exports = function(world) {
+  let browser = world.browser;
+
+  // world.given('I go to the new online DL application page', function(done) {
```
baccigalupi: Here too, we don't need to hold on to commented code

PR #11, opened 2017-08-15
Feature hair color refactor

teresavshen:
- Implements hair color features
- Removed duplication by using arrays for hair color
- Implemented cukes

Next round of refractor can be done after merged with master.


client/containers/summary-handler.jsx
```diff
@@ -18,6 +19,10 @@ const hasResidenceAddressEntered = (props) => {
   return props.street || props.city || props.zip;
 };
 
+const hasHairColorEntered = (props) => {
```
baccigalupi: Hey, I am looking at this and it looks like this check looks for the `hairColor` attribute, and also the props passed in are passing in only that value. Can you confirm that this is really working?

client/helpers/on-button-click.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+export default function onClickGenerator(action, dispatch) {
+  return function onClick(event) {
+		let target  = event.target;
```
baccigalupi: Indentation looks really funky here. Can you check your IDE and make sure it is set to 2 space indentation, and also fix anything in this commit that doesn't match?

client/helpers/on-button-click.js
```diff
@@ -0,0 +1,9 @@
...
+		let value = target.value;
+    dispatch(action(value));
+	};
```
baccigalupi: One more question, I haven't looked at the old onChange handler, but how is this different?

client/presentations/hair-color-form.jsx
```diff
@@ -0,0 +1,36 @@
...
+import alicePath from '../helpers/alice-path';
+
+const HairColorForm = (props) => {
+
+  const hairColorOptions = ['Auburn', 'Bald', 'Black', 'Blonde', 'Brown', 'Gray', 'Red', 'White', 'Other'];
+  let formContent = [];
+
+  hairColorOptions.forEach((option) => {
+    formContent.push(
+      <div key={option} className="input-container hair-color">
```
baccigalupi: I would throw this into a mini-component or function. It will clear up the method and make it easier to read. Also, I think the color options can move out of the Form component since they are constants.

client/presentations/summary-view.jsx
```diff
@@ -30,3 +30,11 @@ export const SummaryResidenceAddress = (props) => {
     </div>
   );
 };
+
+export const SummaryHairColor = (props) => {
+  return (
+    <div className='summary-page'>
+      <p> Hair Color: {props.hairColor.hairColor} </p>
```
baccigalupi: Oh ... this is a weird data organization. I get why the summary check does what it does now. Wondering if there is a better way to build the data. Like does this attribute have to be an object. Can it just be a value?

client/reducers/update-hair-color.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
```
baccigalupi: Same question as above. Can we easily make the data structure simpler since it is just one value?

client/containers/summary-handler.jsx
logicgraphy: It has to be an object with key/value pairs. We can make hairColor as property of a bigger or parent object. We should consider refactoring/compacting reducers to make them generalized  in a way. That way it will have bigger and fewer redux state objects.

client/helpers/on-button-click.js
logicgraphy: Added onClick handler to handle click events for button elements.

baccigalupi:
Hey @teresavshen @logicgraphy I am confused about the state of this PR. Maybe at 11:30 or later?

PR #13, opened 2017-08-16
Sets CA as default state for residence address

logicgraphy:
-Rework on story/feature residence address 
- Adds 'CA' as default state for address
-Summery view shows address only if street, city or zip has values ; no change is required.


logicgraphy:
Made required code change.

PR #15, opened 2017-08-18
Refactor a bit

baccigalupi:
As you can see on our code-climate badge, our code quality is going down. I wanted to take a stab at eliminating some duplication and refactoring. 

This set of commits simplifies a lot about the selector button forms for hair and eye, which are pretty close to the same. Also related to this, there is a lot that can and should be done with the style sheets and guide instead of resorting to custom CSS. 

It also tackles duplication in the actions (partially). It would be great if someone looked into the contacts form and figured out why it can't use a generated action.

Last, it uses the TextInput objects in the NameForm. 

There are a number of other places we can work on cleaner code, but this is a start.


logicgraphy:
Fixed contacts form to use a generated action.

PR #17, opened 2017-08-24
Feature mailing address

logicgraphy:
Implements mailing address entry feature.
Created a AddressTemplate component with common address form structure and modified residential address form accordingly.

### DIFF
--- a/client/helpers/on-input-change.js
+++ b/client/helpers/on-input-change.js
@@ -4,10 +4,21 @@ export default function onChangeGenerator(action, dispatch) {
   return function onChange(event) {
     let target  = event.target;
     let name = target.getAttribute('name');
-    let value = target.value;
-    if( target.getAttribute('type') === 'checkbox') {
-      value = target.checked
+    let value = '';
+
+    switch (target.getAttribute('type')) {
+      case 'text':
+        value = target.value;
+        break;
+
+      case 'checkbox':
+        value = target.checked;
+        break;
+
+      default:
+        value = target.value;
     }
+
     dispatch(action(name, value));
   };
 }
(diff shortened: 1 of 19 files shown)


baccigalupi:
I just looked at the failing codeclimate stuff. A lot of it was miscomprehension of React and duplication, but I thought the summary area duplication was probably a good thing to clean up. I think we probably have an address summary blob that can be reused for both residential and mailing. Can you make that change ... meanwhile I will look at the rest of the code!

client/helpers/on-input-change.js
```diff
@@ -5,6 +5,9 @@ export default function onChangeGenerator(action, dispatch) {
     let target  = event.target;
     let name = target.getAttribute('name');
     let value = target.value;
+    if( target.getAttribute('type') === 'checkbox') {
```
baccigalupi: This works, but seems like we should have a separate input change handler for checkboxes. It would make it more clear. That may mean that the form-change handler will need to add this to the mix and that the input for the checkbox will need to do something like this `<input type='checkbox' onChange={onCheckboxChange} ... >`.

Thoughts?

logicgraphy: Regarding onChange handler; right now we have two input types (text and checkbox), we will have more of these input types in the form (may be radio, email, date etc.). We can create onChange handler for each type and add all of them to the connect-form, or we can just have one handler (like we have now) and put cases/conditions to evaluate value based on the type of input. I think we can go with the later approach for now to keep it simple and if it starts looking ugly, then we can split into individual handlers.

logicgraphy:
Implements requested changes. 
Pending code duplicate issues indicated by code climate will be pushed into another PR.

PR #21, opened 2017-09-06
Adds a first example react unit test!

baccigalupi:
* Uses enzyme and jsdom to render components in test mode
        * Also sets up mocha to transpile the es6/react into plain js

Let's start building unit tests for all the components that we test.
Test the logic, not just attribute rendering!


baccigalupi:
Ok. Looks like we have to hold on this because CI does not act the same as locally. I think there are some kinks to work out!

Will ping you when it is ready

PR #23, opened 2017-09-11
Feature save data

logicgraphy:
@baccigalupi Creating this PR to get feedback on the approach. This is still work in progress and all test cases are not implemented. I hit a blocker where I need to generate actions to update current app state with response from back-end/API. I am looking into redux-thunk and redux-saga to figure out the best possible solution.


client/actions/action-types.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
```
baccigalupi: This change to constants for action types seems good. I would separate it into a separate PR

client/containers/application-reload-button.jsx
```diff
@@ -0,0 +1,32 @@
...
+import React from 'react';
+
+import { connect } from "react-redux";
+import {getUserData}  from "../actions/index";
+
+const ApplicationReloadButton = ({uuid, onApplicationReload}) => {
+  return (
+    <div className='inner-bottom'>
+      <button name='reloadApplication'
+        onClick={ () => onApplicationReload(uuid) }
```
baccigalupi: let's separate these kind of inline anonymous function into named variables. Makes it easier to read and name in a way that will be understandable.

client/containers/eye-color-form-container.jsx
```diff
@@ -3,7 +3,7 @@
 import React from 'react';
 
 import colorFormBuilder     from '../presentations/color-form-builder';
-import { updateEyeColor }   from "../actions/index";
+import { updatePhysicalAttributes }   from "../actions/index";
```
baccigalupi: This kind of refactor is a good idea too. I would break this off from this PR and put it into a separate PR pre-this one that is refactor prep. Similar to the changes to the constant action types

client/helpers/data-services.js
```diff
@@ -0,0 +1,64 @@
+import request from 'superagent'
+
+const dataService = store => next => action => {
+
```
baccigalupi: I would want to pull all these case blocks into separate functions, and maybe into separate modules. TESTS! Separate modules means that they can be tested independently.

client/presentations/home.jsx
```diff
@@ -12,7 +12,7 @@ const Home = () => {
         <Link to={ alicePath('/summary') }>summary</Link>
       </li>
       <li>
-        <Link to={ alicePath('/about-me/names') }>about-me-names</Link>
+        <Link to={ alicePath('/about-me/names') }>names</Link>
```
baccigalupi: This was already changed. So, let's drop it.

db/knex.js
```diff
@@ -0,0 +1,5 @@
+'use strict'
+
+var environment   = process.env.NODE_ENV || 'development';
+var config        = require('../knexfile.js')[environment];
+module.exports = require('knex')(config);
```
baccigalupi: It might be good to put this into a closure: 

```
const knex = require('knex');
module.exports = () => { knex(config); }
```

That way the loading of the file does not have side effects (which leads to complex bugs and problems. If we want to store this somewhere on the application level as a singleton type of a structure, then creating a value object with the connection and attaching it to a central place in the app would be better.

knexfile.js
```diff
@@ -0,0 +1,16 @@
+// Update with your config settings.
+
+module.exports = {
+
+    development: {
+      client: 'pg',
+      connection: 'postgres://localhost/alice_dev',
```
baccigalupi: This needs to be in an environmental variable so that we can move this to staging, production etc. Talk to Dave1 or me about how to do this abstraction for databases in node.

server.js
```diff
@@ -39,4 +39,50 @@ server.get('/protected', authenticate, function(req, res){
   res.json({hello: 'protected world'});
 });
 
+//TODO: Move to router module
+
+server.get('/api/v1/user-data/:uuid', function(req, res){
+  knex.select().from('Application').where('id', req.params.uuid)
```
baccigalupi: Let's put each of these functions with req, res into its own controller module, tested.

PR #28, opened 2017-09-12
Changes the data organization:

baccigalupi:
We need to start storing information about UX and/or fetch state.
Therefore, this commit reorgs the existing data in the `application`
key namespace.

It uses combine reducers at the root level and also at the application
namespace.

Seems to work!


logicgraphy:
Yep, seems to work! And pretty smart way to put states into application scope. Will be much easier to save and reload application.

PR #31, opened 2017-09-18
Apis crud application

LeDeep:
Create get and post API endpoints for user data. Added unit tests to test the request handlers for the APIs.

Needed to setup postgres instance to allow TravisCI to connect to the database to run the build. Ran into some issues configuring the environment variables. Hence the numerous commits. Ultimately, we needed to make sure our scripts in the .travis.yml file were running in the correct asynchronous order.


package.json
```diff
@@ -29,6 +29,7 @@
     "babel-preset-react": "^6.24.1",
     "babel-register": "^6.24.1",
     "chai": "^4.0.2",
+    "chai-http": "^3.0.0",
```
baccigalupi: Is this being used anymore?

server.js
```diff
@@ -10,6 +10,7 @@ const passport    = require('passport');
 const helmet      = require('helmet');
 const jwtStrategy = require('./server/config/jwt-strategy').strategy;
 const logging     = require('./server/config/logging');
+const api_v1      = require('./server/api/v1');
```
baccigalupi: We are creating an API that is used just for our client side application. Let's not version it. Versioning is one of those "optimization in advance" things that complicates all the code going forward. It is usually really easy to add versioning when your really need it.

server/api/v1/index.js
```diff
@@ -0,0 +1,88 @@
...
+  knex('applications')
+  .join('addresses', 'applications.id', 'addresses.resident_id')
+  .join('emails', 'emails.resident_id', 'applications.id')
+  .join('phone_numbers', 'phone_numbers.resident_id', 'applications.id')
+  .select(...APPLICATIONS_TABLE_COLUMNS.map((d) => { return 'applications.' + d }),
+          ...ADDRESSES_TABLE_COLUMNS.map((d) => { return 'addresses.' + d }),
+          ...EMAILS_TABLE_COLUMNS.map((d) => { return 'emails.' + d }),
+          ...PHONE_NUMBERS_TABLE_COLUMNS.map((d) => { return 'phone_numbers.' + d }))
+  .where('applications.id', req.params.uuid)
+    .then(function(data){
```
baccigalupi: This little part of the function is the actual controller that belongs in here. The rest of this is a model that should be extracted and tested independently.

Wondering why the select fields were necessary, couldn't we select '*'?

server/api/v1/index.js
```diff
@@ -0,0 +1,88 @@
...
+      res.send(data);
+    });
+}
+
+function postUserDataHandler(req, res){
+  knex('applications')
+  .insert({
+    id: req.body.uuid,
+    type: req.body.type,
```
baccigalupi: the `type`, `source`, and `number` aren't yet known from the client. Same with `name_suffix`. Probably we shouldn't have added these fields to the db schema until we needed them ... That is totally my fault for copying and pasting without really thinking about it.

server/api/v1/index.js
```diff
@@ -0,0 +1,88 @@
...
+    first_name: req.body.first_name,
+    middle_name: req.body.middle_name,
+    last_name: req.body.last_name,
+    name_suffix: req.body.name_suffix,
+    date_of_birth: req.body.date_of_birth,
+    language: req.body.language,
+    hair_color: req.body.hair_color,
+    eye_color: req.body.eye_color
+  })
+  .then(() => {
```
baccigalupi: This should be broken into separate functions, and extracted out of the controller realm too.

server/api/v1/index.js
```diff
@@ -0,0 +1,88 @@
...
+  .then(()=> {
+    return knex('phone_numbers')
+      .insert({
+        resident_id: req.body.uuid,
+        number: req.body.phone_number,
+      })
+  })
+  .then(() => {
+    return knex('applications')
+    .join('addresses', 'applications.id', 'addresses.resident_id')
```
baccigalupi: This is the first place you would get to use that same function used up in the GET request if it was extracted and separate from the small controller part.

test/server/apis/user-data-save-v1-unit-test.js
```diff
@@ -0,0 +1,195 @@
+'use strict';
+
+// Set APP_ENV to 'test' to connect with test db
+process.env.APP_ENV = 'test';
+
+const expect = require('chai').expect;
```
baccigalupi: Can we stick with `assert` on this. I am open to a conversation. This looks a lot better in the test, but the error messages when something goes wrong are really cryptic. Things about undefined not having properties/functions instead of just a message about why the test failed.

.travis.yml
```diff
@@ -7,7 +7,13 @@ cache:
...
   - JWT_SECRET=s3kret
 script:
+  - node_modules/knex/bin/cli.js migrate:latest
   - npm run test:coverage
 after_success:
   - npm install -g codeclimate-test-reporter
   - codeclimate-test-reporter < coverage/lcov.info
+services:
+  - postgresql
+before_script:
```
baccigalupi: Configuring CI is rough! And usually does involve some trial and error in commits.

How do these integrate with the knexfile? The pattern in the app uses either a `DATABASE_URL` in an env var, or the package.json. Wondering how this integrates without changes to that pattern. Is it in a different commit?

logicgraphy: This worked because I have had DATABASE_URL environment variable added in travis to the repository settings. I am moving it to travis.yml so that everything is pre-configured.

package.json
LeDeep: No. Will remove.

server.js
LeDeep: Makes sense. Will remove v1.

test/server/apis/user-data-save-v1-unit-test.js
logicgraphy: Done! Removed chai dependency and replaced expect with assert

.travis.yml
baccigalupi: Ok. got it! Thanks

server/api/v1/index.js
logicgraphy: Extracted the database operations into a separate user-data service component.
Select fields are used to assign alias in few case. For example phone_number and email_adress.

.travis.yml
```diff
@@ -5,9 +5,15 @@ cache:
   directories:
     - "node_modules"
 env:
-  - JWT_SECRET=s3kret
+  - JWT_SECRET=jot-it-down JWT_USERS='["SOS"]' DATABASE_URL=postgresql://ci_tester:pa33word@localhost/travis_ci_test
```
baccigalupi: Now that I have that commit on master, this will probably need to change.

PR #32, opened 2017-09-20
Eye color form navigates to hair color form

baccigalupi:
* Uses the new select buttons
        * Simplifies and concatenates the scenarios for faster run times
        * No new React unit tests, because all presentation and
          configuration of existing
        * Extracts helper function in navigation steps


baccigalupi:
In general, I want us to pay attention to code climate. I asked them how to tune the kind of duplication we are seeing in React from good extraction of duplicate code, and they were really useless on a support level. 

I looked at the duplication and right now it doesn't look like the right time to find a way to clean it up. One of the things I am thinking about with these select buttons is that images are in future features. If we do too much abstraction now, that logic will be really hard to unpack.

Anyway, this is all to say that I looked at the CodeClimate issues and don't agree that they are real issues. Let me know what you think when you review.

PR #40, opened 2017-09-25
[finishes #150524761] Enter social security number page

baccigalupi:
* Adds action, reducer, container and presentation
        * Refactors summary views to render null element when data not
          present
        * Adds data present test for application

The refactor on summary is only for the elements that were on master. Address stuff still needs to be merged.

### DIFF
--- a/client/containers/summary-handler.jsx
+++ b/client/containers/summary-handler.jsx
@@ -5,17 +5,21 @@ import { connect } from 'react-redux';
 
 import HomeLink from '../presentations/home-link.jsx';
 import {
-  SummaryEmpty,
-  SummaryNames,
   SummaryHomeAddress,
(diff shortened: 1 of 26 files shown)


client/containers/summary-handler.jsx
```diff
@@ -26,47 +30,30 @@ const hasContactDetailsEntered = (props) => {
...
-  }
-
-  if (dataPresent.value(props.weight)) {
-    contents.push(<SummaryWeight weight={props.weight} key='weight'/>);
-  }
-
-  if (!contents.length) {
-    contents.push(<SummaryEmpty key='summary'/>);
-  }
+  contents = contents.reduce((summaries, item) => {
```
baccigalupi: Oops. this shouldn't be here

PR #41, opened 2017-09-26
Cleans up the HomePage for CodeClimate

baccigalupi:
The home page was showing up as having high duplication, and a lot of
churn in code climate. This commit reduces duplication, and allows us to
add links by just adding the description and path.


baccigalupi:
And of course, CodeClimate is seeing weird duplication between jsx templates that doesn't really exist. Anyway, this is a good cleanup

LeDeep:
Looks Good! I really like this refactor.

PR #42, opened 2017-09-26
Implement mailing address form

LeDeep:
- Implement mailing address form into the flow
- Add acceptance tests
- Will likely need to refactor the reducer to incorporate logic for mailing address being same/different as residential address (see story #150495457)


client/presentations/summary-view.jsx
```diff
@@ -23,6 +23,15 @@ export const SummaryHomeAddress = (props) => {
...
 
+export const SummaryMailingAddress = (props) => {
+  return (
+    <div className='summary-section'>
+      <p>Mailing address: </p>
+        <SummaryAddressBlob address={ props.mailingAddress }/>
+    </div>
+  );
+};
+
```
logicgraphy: This can be refactored following similar approach as of rest of the summary components.

LeDeep:
Updated home and mailing addresses to follow the summary component pattern. Since contact details will need to be refactored anyway, I left ContactDetails in the summary_view for now. Figure we can refactor that portion all at once.

PR #45, opened 2017-09-27
Implements voter preferences intro page

teresavshen:
- Implements intro page for voter preferences


client/presentations/voter/voter-preferences-intro-form.jsx
```diff
@@ -0,0 +1,33 @@
...
+      <p>You can become a member of a political party or change your party membership.</p>
+      <h4>Vote by mail</h4>
+      <p>Choose if you want to get your ballot by mail.<br />
+      You can always change your mind and vote at a polling place.</p>
+      <h4>Language</h4>
+      <p>Choose what language you would like to get your voting materials in.</p>
+      <h4>Contact Information</h4>
+      <p>How would you like election campaigns to reach you — if at all?</p>
+
+      <form onSubmit={ props.onSubmit } className='voter-preferences-intro-form' >
```
baccigalupi: I think this should just be an `<a>` link. like the ones on the home page. Much easier to implement, and it feels confusing that this is a form with no real fields.

test/features/step_definitions/voter-preference-intro-steps.js
```diff
@@ -0,0 +1,27 @@
...
+const assert = require('assert');
+
+module.exports = function(world) {
+  let browser = world.browser;
+
+  world.then('I see text for voter preference intro', function(done){
+    browser
+    .text()
+    .then((text) => {
+      assert.ok(text.includes('We will now ask for your preferences on the following voter registration details:'), 'voter preference intro page text 1 is incorrect');
```
baccigalupi: We should avoid testing exact content. It is likely to change over the course of the application development, and testing to this level will make our tests very brittle. What we really want to check is something in the template that doesn't change, but assures that the page was rendered without error. A selector is usually fine.

In general, we want to test logic, not configuration (or content).

PR #46, opened 2017-09-28
Implement voter registration eligibility form page

LeDeep:
- Add page allowing users to select if they meet voter eligibility requirements
- Add acceptance tests


test/features/voter-registration-eligibility.feature
```diff
@@ -0,0 +1,79 @@
+Feature: Check Voter Registration Eligibility
+  As a DMV Customer
+  I want to check if I qualify for voter registration status
+  so that I can register to vote during this DMV transaction
+
+  Scenario: Seeing the empty form
```
baccigalupi: Can you try wrapping these up into one or two scenarios. The product specification is in cucumber, but we own and maintain the test suite. Each scenario has a start/stop cost in time in the suite, and also, the sheer number of scenarios gets confusing sometimes.

You can start to group given/when/then blocks within a scenario. I compressed the scenarios for entering a name, dob etc if you are looking for an example.

client/containers/voter/eligibility-requirements-form-container.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+import React from 'react';
+
+import { updateEligibilityRequirements }   from '../../actions/index';
+import Form                           from '../../presentations/voter/eligibility-requirements-form.jsx';
+import connectForm                    from '../../helpers/connect-form';
+import navigateOnSubmit               from '../../helpers/navigate-on-submit';
+
+const ConnectedForm = (props) => {
+  console.log(props);
```
baccigalupi: This should go!

client/containers/voter/eligibility-requirements-form-container.jsx
```diff
@@ -0,0 +1,38 @@
...
+  const continueDisabled = false;
+  let value = props.eligibilityRequirements;
+  let onSubmit = navigateOnSubmit('/summary', props);
+
+  if(value === 'Yes' || value === 'No') {
+    onSubmit = navigateOnSubmit('/about-me/voter/opt-out', props);
+  } else {
+    onSubmit;
```
baccigalupi: might make more sense to either set both paths in the if/else or just not have the else block since you are using the default value.

client/presentations/summary/eligibility-requirements.jsx
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+import React from 'react';
+import * as dataPresent from '../../helpers/data-present';
+
```
baccigalupi: Might be good to get some unit tests for this. There is a bit of logic that would be good to test here instead of in acceptance tests.

package.json
```diff
@@ -61,12 +61,14 @@
     "helmet": "^3.6.1",
     "jsonwebtoken": "^7.4.1",
     "morgan": "^1.8.2",
+    "npm": "^5.4.2",
```
baccigalupi: Why do we need `npm` and `test`? Is there something about your dev setup, or does this allow easier commands ?

test/features/voter-registration-eligibility.feature
LeDeep: Got it down to 4 scenarios. Since the continue button is enabled, I wanted to make sure I captured that behavior.

client/containers/voter/eligibility-requirements-form-container.jsx
LeDeep: oops!

client/containers/voter/eligibility-requirements-form-container.jsx
LeDeep: Removed the else. I think it reads better, and like you said, the default value is set.

client/presentations/summary/eligibility-requirements.jsx
LeDeep: I'll go ahead and make a separate chore for the tests.

package.json
LeDeep: Not sure what happened. It must have been my local setup. I removed them from the file.

PR #48, opened 2017-09-29
Political party choose

LeDeep:
- Allow users the option to add or update a political party preference

### DIFF
--- a/client/containers/voter/political-party-choose-form-container.jsx
+++ b/client/containers/voter/political-party-choose-form-container.jsx
@@ -9,7 +9,6 @@ import navigateOnSubmit               from '../../helpers/navigate-on-submit';
 import * as dataPresent               from '../../helpers/data-present';
 
 const ConnectedForm = (props) => {
-  console.log(props);
   let value = props.politicalPartyChoose;
   const continueDisabled = !dataPresent.value(value);
   let onSubmit = navigateOnSubmit('/about-me/voter/ballot-by-mail', props);
--- a/client/containers/voter/political-party-choose-form-container.jsx
+++ b/client/containers/voter/political-party-choose-form-container.jsx
@@ -1,0 +1,39 @@
+'use strict';
+
+import React from 'react';
+
+import { updatePoliticalPartyChoose }   from '../../actions/index';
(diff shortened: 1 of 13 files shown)


client/containers/voter/political-party-choose-form-container.jsx
```diff
@@ -0,0 +1,39 @@
...
+  let value = props.politicalPartyChoose;
+  const continueDisabled = !dataPresent.value(value);
+  let onSubmit = navigateOnSubmit('/about-me/voter/ballot-by-mail', props);
+
+  if(value === 'Yes') {
+    onSubmit = navigateOnSubmit('/about-me/voter/political-party', props);
+  } else {
+    onSubmit;
+  };
+
```
LeDeep: @baccigalupi any suggestions for how to write a unit test for this? It's similar to my previous commit, and we have this logic in a few other places. I haven't been able to get a good test for this. Curious if you had any ideas.

baccigalupi: In git history there are some tests for presentation compontents that navigated. When we moved that logic to the container level I realized it was just configuration.

This is really logic though, that should be unit tested. So, the basic process is to `mount` the component with enzyme. As soon as you do that you are going to be told that the routing portions are an issue. At that point I imported a memory router and wrapped the component I was mounting. 

It was rough to figure out, but if you look through our git history on github or on the CLI you can find some examples. Once you have the pattern it is pretty easy to test.

client/containers/voter/political-party-choose-form-container.jsx
```diff
@@ -0,0 +1,39 @@
...
+import React from 'react';
+
+import { updatePoliticalPartyChoose }   from '../../actions/index';
+import Form                           from '../../presentations/voter/political-party-choose-form.jsx';
+import connectForm                    from '../../helpers/connect-form';
+import navigateOnSubmit               from '../../helpers/navigate-on-submit';
+import * as dataPresent               from '../../helpers/data-present';
+
+const ConnectedForm = (props) => {
+  console.log(props);
```
baccigalupi: Oop. Another log!

PR #57, opened 2017-10-04
Implemented opt-out choice for voter registration

teresavshen:
- Implemented Voter Registration Opt-out page
- Reroutes to Voter Preferences Info page or Summary page depending on user selection


test/features/step_definitions/shared-form-steps.js
```diff
@@ -115,4 +114,38 @@ module.exports = function (world) {
     .then(() => { done(); })
     .catch(done);
   });
+
```
logicgraphy: Move these feature specific tests out from shared steps.

PR #65, opened 2017-10-09
Page title + code refactor

logicgraphy:
- Implements feature to update document/page title based on navigation
- Moves files related to DL application to 'motor' folder
- Some code beautification.


client/containers/motor/date-of-birth-form-container.jsx
```diff
@@ -0,0 +1,33 @@
+'use strict';
```
baccigalupi: I think putting these into a folder to organize them is a GREAT idea. Thanks for doing that. I think the domain area for this is an `application` which is overused in the web dev space, so maybe `apply` would be good as a folder name.

PR #74, opened 2017-10-12
Restyling

baccigalupi:
In prep for a little stakeholder release this week, this is a little bit of styling to get us closer to the developer intent. 

Since there is going to be other restyling coming up and we are using a very particular style guide, I wanted everyone to check out the diff on this. Very little per-page customization was done. Instead a bunch of scss sheets that change everything were changed, treating the css as though it were a programmed system. This is the way future changes have to happen too.


LeDeep:
Fancy! Looks real nice.

PR #80, opened 2017-10-12
Update the contact choice page

teresavshen:
- Updated contact choice page
- Renamed 'contact method' to 'contact choice' to reduce confusion


client/containers/routes.jsx
```diff
@@ -68,7 +68,7 @@ class Router extends React.Component {
         <Route path={ alicePath('/about-me/voter/political-party') } component={PoliticalPartyPreference} />
         <Route path={ alicePath('/about-me/voter/ballot-language') } component={BallotLanguage} />
         <Route path={ alicePath('/about-me/voter/ballot-by-mail') } component={BallotByMail} />
-        <Route path={ alicePath('/about-me/voter/contact-choice/') } component={ContactMethods} />
+        <Route path={ alicePath('/about-me/voter/contact-choice/') } component={ContactChoice} />
         <Route path={ alicePath('/about-me/voter/email-phone') } component={ContactDetails} />
```
baccigalupi: Is ContactDetails and the /email-phone path still being used? If not, would be a good time to delete all the code.

test/features/step_definitions/navigation-steps.js
```diff
@@ -239,6 +235,10 @@ module.exports = function(world) {
     clickAndWaitForPage('a.success-visit', '.success-visit-info', done);
   });
 
+    world.when('I visit voter opt out page', function(done) {
```
baccigalupi: It looks like the indentation is wrong here. If so, can you fix it?

client/containers/routes.jsx
teresavshen: The /email-phone path is being use when 'Yes' is selected on Contact Choice. I'll remove /email-phone-remove, which is no longer used.

PR #87, opened 2017-10-13
Browser history navigation

logicgraphy:
- Replaces BrowserRouter with Router, a lower level react-router component which gives access to history object explicitly. 
- Fixes acceptance test navigation to use history.push(${path}) in order to end dependency of clicking on links for navigation.
- The dependency of clicking on links for navigation still exists at few places within step definitions for acceptance tests, these will be addressed ones the links are completely removed.


client/containers/app.jsx
```diff
@@ -1,24 +1,31 @@
...
+import { Provider }                 from "react-redux";
+import { createStore }              from "redux";
+import createBrowserHistory         from 'history/createBrowserHistory'
 
-import reducers from "../reducers";
-import Routes from './routes.jsx';
+import reducers                     from "../reducers";
+import Routes                       from './routes.jsx';
+
+const history = createBrowserHistory()
```
baccigalupi: This all looks great. Thanks for figuring out a clean way to create this back door!

PR #101, opened 2017-10-18
Previous names selection

LeDeep:


LeDeep:
@logicgraphy Refactored to use one reducer, and added specs. Let me know what you think.

PR #108, opened 2017-10-20
Bug: Buttons should be the same size

teresavshen:
- CSS updates to make buttons the same size
- Alignment adjustments for various screen sizes


LeDeep:
It looks like this is working on some pages, but not on others. For example, the buttons look good on the 'voter/eligibility-requirements' and '/organ' pages. But they are still different sizes on the 'license-history' and 'donate-contribution' pages.

Also, on a few of the pages, the height of the buttons are different, possibly because of the width changes. For example, 'political-party-choose' and 'political-party' pages.

PR #109, opened 2017-10-24
Combines contact choice and contact details into single political contact form

logicgraphy:
- Removes contact choice form and container
- Removes contact details form and container
- Adds political contact container form as a controller for contact choice and details.


baccigalupi:
Hey, noticed the CodeClimate shows duplication on the email etc. Do we need both forms of the contact for what is visible.

client/helpers/data-present.js
```diff
@@ -59,6 +53,18 @@ const previousNamesInfo = (props) => {
   );
 }
 
+const politicalContact = (props) => {
+  return props && (
+    value(props.emailAddress) || value(props.phoneNumber) || value(props.shouldContact)
+  );
+}
+
```
baccigalupi: These look like the same thing. Probably we want one function named well enough that it makes sense in each context!

client/presentations/summary/political-contact.jsx
```diff
@@ -0,0 +1,31 @@
...
+  if(shouldContact !== 'Yes') {
+    if(shouldContact === 'Skip Question') { shouldContact = 'No Answer'; }
+    return (
+      <div className='summary-section'>
+        <p> Should Contact: {shouldContact} </p>
+      </div>
+    );
+  }
+
+  //if (!dataPresent.contactDetails(props.politicalContact)) { return null; }
```
baccigalupi: Oh this is that null return I was talking about earlier! Anyway, let's get this comment out if we don't need it.

client/actions/index.js
```diff
@@ -32,16 +31,15 @@ export const TYPES = {
   UPDATE_POLITICAL_PARTY_CHOOSE:          'UPDATE_POLITICAL_PARTY_CHOOSE',
   UPDATE_POLITICAL_PARTY_PREFERENCE:      'UPDATE_POLITICAL_PARTY_PREFERENCE',
   UPDATE_BALLOT_LANGUAGE:                 'UPDATE_BALLOT_LANGUAGE',
-  UPDATE_CONTACT_CHOICE:                  'UPDATE_CONTACT_CHOICE',
-  UPDATE_OPT_OUT:                         'UPDATE_OPT_OUT'
+  UPDATE_OPT_OUT:                         'UPDATE_OPT_OUT',
```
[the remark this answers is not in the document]
logicgraphy: We combined action and reducer for 'contact choice' and 'contact details' into one called 'political-contact'

PR #111, opened 2017-10-30
Organ donation

LeDeep:
- Refactored selector-button to use more page specific id and label, since we now will have pages, with one form that includes the same value multiple times


client/helpers/data-present.js
```diff
@@ -24,6 +24,12 @@ const height = (props) => {
   return props && value(props.feet);
 }
 
+const organDonation = (props) => {
+  return props && (
+    value(props.donate) && value(props.contribute)
```
logicgraphy: this to be updated to use new hasAllAttributes method

PR #112, opened 2017-10-31
Api refactor

logicgraphy:
- Implements data parsers to convert data between existing client format and DB.
- Extracts a bit of control logic into controller module within server.
- Fixes api and model level tests.


server/controllers/index.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const clientParser    = require('../models/client-data-parser');
+const serverParser    = require('../models/server-data-parser');
+const getAppModel     = require('../models/db/get-application');
+const createAppModel  = require('../models/db/create-application');
+
```
baccigalupi: I love these controllers in a different directory!

Seems like we need to handle the error case on these, which should be easier to test now that it is separate.

server/models/server-data-parser.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
```
baccigalupi: Is this a serializer? Meaning it is taking on the role of converting the data back to the client in a way that it can understand?

baccigalupi: Also can you check out the CodeClimate and see what you think about the errors they are reporting?

PR #119, opened 2017-11-02
Combine height and weight questions

teresavshen:
- Combines height and weight questions to individual traits page


client/presentations/apply/traits-height-weight-form.jsx
```diff
@@ -4,11 +4,12 @@ import React            from 'react';
 import HomeLink         from '../home-link.jsx';
 import NumberInput      from '../number-input.jsx';
 import ContinueButton   from '../continue-button.jsx';
+import navigateOnSubmit from '../../helpers/navigate-on-submit';
```
logicgraphy: Not sure if we need navigateOnSubmit in presentation form.

client/reducers/application/update-traits-height-weight.js
```diff
@@ -6,8 +6,9 @@ import formObjectReducer from './form-object-reducer';
 function defaultState() {
   return {
     feet: '',
-    inches: ''
+    inches: '',
```
logicgraphy: I think we should have a height object to enclose feet and inches.
For e.g. 
return {
height: { feet: ' ', inches: ' ' },
weight: ' ' 
} 
And then, will have to fix other places for these changes.

logicgraphy: I think we should have a height object to enclose feet and inches.
For e.g.
return {
height: { feet: ' ', inches: ' ' },
weight: ' '
}
And then, will have to fix other places for these changes.

test/client/helpers/data-present-test.js
```diff
@@ -67,14 +67,19 @@ describe('dataPresent', function() {
...
+        'date not present with all fields'
+      );
     });
 
-    it('is false without the feet', function() {
-      assert(!dataPresent.height({inches: '5'}), 'height data present with inches only');
-      assert(!dataPresent.height({}), 'height data present when blank');
+    it('is false when only partial data present', function() {
+      assert(
+        !dataPresent.date({inches: '5', weight: '201'}),
```
logicgraphy: date() should not be here. 
Maybe test requires to check height has 'feet' value and not only 'inches'

PR #126, opened 2017-11-09
Server/API integration baked into client app

logicgraphy:
- Implements ReduxThunk middleware and axios for creating async actions and calling APIs.
- Adds new and separate files for actions and reducers to handle API related operations.
- Fixes test cases. Adds a wait to tests where API call is made because no other workaround is available in node-horseman
- Additional required tables and fields will be added in separate PRs.
- Build configuration to inject API host for Heroku (dev & staging) will be covered in a separate chore.


test/features/step_definitions/navigation-steps.js
```diff
@@ -28,6 +28,12 @@ module.exports = function(world) {
 
   world.given('I go to the new online DL application', function(done) {
     browser
+      .on('consoleMessage', function( msg ){
```
baccigalupi: These seem like really good debugging helpers!

webpack.config.js
```diff
@@ -1,7 +1,25 @@
 'use strict';
 
-const path              = require('path');
-const ExtractTextPlugin = require("extract-text-webpack-plugin");
+const path                = require('path');
+const webpack             = require('webpack');
+const ExtractTextPlugin   = require("extract-text-webpack-plugin");
+let apiHost;
+
```
baccigalupi: Do we need to have a full url, can we just use a path? For example with the `fetch` polyfill, you can just pass it the path and it knows to use the same domain as the one you are using. If we can avoid putting this host information in here it would be good.

PR #139, opened 2017-11-20
 Hiding the brand new save and reload buttons

logicgraphy:
- Adds a APP_ENV variable into client app which is assigned value during webpack build.
- Adds conditional rendering for 'Save' & 'Reload' buttons for development and test environments only.
- Adds app env var to travis CI configuration for tests.
- Also, added appropriate APP_ENV variables for corresponding Heroku instances.

### DIFF
--- a/client/containers/summary-handler.jsx
+++ b/client/containers/summary-handler.jsx
@@ -76,17 +76,24 @@ const SummaryHandler = (props) => {
     props.dispatch(postData(data));
   }
 
+  if(APP_ENV === 'development') {
+    contents.push(
+      <div key='save-reload-data'>
+        <div className=' unit relative'>
+          <button id='reloadData' key='reload-data' onClick={loadData}> Reload </button>
+        </div>
+        <div className='unit spacer'></div>
+        <div className=' unit relative'>
+          <button id='saveData' key='save-data' onClick={saveData}> Save </button>
+        </div>
+      </div>
+    );
+  }
+
   return (
     <div className='summary'>
       <HomeLink />
       { contents }
-      <div className=' unit relative'>
-        <button id='reloadData' key='reload-data' onClick={loadData}> Reload </button>
-      </div>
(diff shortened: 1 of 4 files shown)


server/config/env.js
```diff
@@ -3,7 +3,7 @@
 const dotenv = require('dotenv');
 const env    = process.env.APP_ENV || 'development';
 
-if (env === 'development' || env === 'test') {
+if (env === 'development' || env === 'test' || env === 'staging') {
```
baccigalupi: In the heroku environments we shouldn't be using dotenv to load variables. Instead we want all those actually set as environmental variables. Might be that we need to set the `dlite-web-dev` heroku to have a different environment like `development-acceptance`, or something just so we can make sure dotenv isn't loading variables in Heroku

webpack.config.js
```diff
@@ -34,7 +34,10 @@ let config = {
     extensions: ['.js', '.json']
   },
   plugins: [
-    new ExtractTextPlugin('app.css')
+    new ExtractTextPlugin('app.css'),
+    new webpack.DefinePlugin({
```
baccigalupi: I am wondering if we need this in webpack ... which is really hard to configure just right and breaks often. I think in `client.js` we will know what the process.env.APP_ENV is and can find a way to store it. Or maybe we can find a way to add this to the redux store when compile `app.js`.

In general, I would prefer we avoid adding anything here unless we have to actually compile a new file or an existing file in a new way.

client/containers/summary-handler.jsx
```diff
@@ -76,17 +76,24 @@ const SummaryHandler = (props) => {
     props.dispatch(postData(data));
   }
 
+  if(APP_ENV === 'development' || APP_ENV === 'test') {
```
baccigalupi: I think this check is going to happen in two places, maybe the logic can move into a helper?

PR #143, opened 2017-11-21
Under 16 voting options

teresavshen:


client/helpers/calculate-age.js
```diff
@@ -0,0 +1,11 @@
+export default (year, month, day) => {
```
baccigalupi: I would want to see really exhaustive unit tests on this. Particularly on the boundary days.

client/reducers/application/update-date-of-birth.js
```diff
@@ -7,7 +7,8 @@ function defaultState() {
   return {
     month: '',
     day: '',
-    year: ''
+    year: '',
+    age: ''
```
baccigalupi: Since this can be calculated at any time, in any container, do we need to store it in state. When we talked about this in pointing, I started to worry that an application that was open for a while would have an outdated version of this age calculation, if we weren't calculating it regularly.

test/features/step_definitions/enter-organ-donation-steps.js
```diff
@@ -176,4 +176,22 @@ module.exports = function(world) {
     .then(() => { done(); })
     .catch(done);
   });
+
+  world.when('I am under 16 years old', function(done) {
+    browser
+      .type('#month', '2')
```
baccigalupi: For longevity ... who knows how long this application will live, this test value for birthday should be calculated based on the current date. The greater age shouldn't be an issue.

teresavshen:
Work in progress. Will create a new PR.

PR #144, opened 2017-11-22
Gets a bit on the visual new styles into the css

baccigalupi:
The designer found our `/styles.html` page on the staging site pretty useful for tinkering with styles. This set of commits does a lot of cleanup on the custom CSS that has been added over the last couple months, then it adds what is possible of the styling from Jacqueline's latest work. In a more terse way of saying it ...

* Adds a lot of input and button work to the styles.html page
* Switches out the fonts and colors to be close to designer intent
* Adjusts the working app to look better under these styles, removing a lot of unnecessary css gunk
* Extracts a LinkButton class for links that look like buttons

### DIFF
--- a/client/presentations/apply/intro.jsx
+++ b/client/presentations/apply/intro.jsx
@@ -1,9 +1,9 @@
 'use strict';
 
 import React        from 'react';
-import { Link }     from 'react-router-dom';
 
 import HomeLink                 from '../../presentations/home-link.jsx';
+import LinkButton               from '../../presentations/link-button.jsx';
 import alicePath                from '../../helpers/alice-path';
 
 const Intro = (props) => {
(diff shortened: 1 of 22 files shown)


baccigalupi:
@kyliepace: Omg. Great interpretation of my sloppiness! Also just good catches!

I fixed the issues with the radio button group, and kind-a fixed the ballot language. Jacqueline is gutting our tappable buttons and changing them into a set of stacked super-styled radios and checks. So, I don't want to invest too much time in getting the current state just right.

In this case it is probably necessary to load up the page to understand how the css impacts the styling, but in general, you don't have to worry about whether the code actually implements the feature. Devs get to just focus on code quality and maintainability when they are doing a review, and also just knowledge sharing. The product team will accept or reject them on the level of 'does this do what I asked'.

PR #148, opened 2017-11-27
Refactor for vizd navigation work

baccigalupi:
Hi team, this is the first step of refactoring the duplication and right abstractions out of the React world. You can opt to use and continue this, or refactor a different way.

#### The thing I was noticing is that there are missing abstractions:

* The header should be it's own component. Also it has UI state that should be in redux, separate from the application state. This commit moves those bits of duplication into a component, but does not do the data or dispatch work.
* A related notion that is missing in a page! Each page will need to dispatch the UI event so that the header is consistent and correct. Also each page currently defines the header/title content, whereas the other components are parts of that page.
* Button groups: This was the motivation for this work. It should be pretty easy to swap out styling and markup for these buttons, but they were copy/pasted into a number of pages. I have noticed that there are link buttons back and forth, and there are also form buttons back and forth. These are two missing components. A bit of work happened here to get these in place, but more needs to happen. Also related to this the back button is pretty high maintenance; I think we are able to do `history.back` in most cases instead of explicitly holding the previous path.

I did this last week, before a holiday weekend. So I am missing some of the intent in this big set of commits! Sorry


client/containers/apply/medical-history-form-container.jsx
```diff
@@ -4,6 +4,7 @@ import React from 'react';
 
 import { updateMedicalHistory }         from '../../actions/index';
 import HomeLink                         from '../../presentations/home-link.jsx';
+import SectionHeader                    from '../../presentations/home-link.jsx';
```
LeDeep: I think this should be '../../presentations/section-header.jsx'

LeDeep:
@baccigalupi looks good. Couple of merge conflicts, so once cleared I can merge with master.

client/containers/apply/medical-history-form-container.jsx
baccigalupi: Oh crap! Copy-paste-disaster

baccigalupi:
@LeDeep I updated this based on your review and also rebased. Please merge if you feel good about it now.

PR #157, opened 2017-12-05
Pre-register to vote

teresavshen:
- Swapped out headers for pre-registration voting pages
- Combined 'voter preferences' and 'voter preferences updated'


client/presentations/voter/contact-methods-choice.jsx
```diff
@@ -9,12 +9,24 @@ const VALUES = ['Yes', 'No', 'Skip Question'];
 let pageTitle = 'DMV: License application - Voting registration'
 
 const ContactChoice = (props) => {
+  let documentHeader = [];
+
+  if ((props.dateOfBirth.age === 16) || (props.dateOfBirth.age === 17)) {
```
kyliepace: I just added to the age calculation in a separate PR so that we can account for fractions of years i.e. someone who is 17.5. How do we want this contact-methods-choice section to handle those cases?

client/containers/voter/choose-party-form-container.jsx
```diff
@@ -19,8 +19,6 @@ const ConnectedForm = (props) => {
   let onBack;
 
   if(props.optOut == "I am already registered to vote in California"){
-     onBack = navigateOnBack('/voting-registration/updating-preferences', props);
-  } else {
      onBack = navigateOnBack('/voting-registration/preferences', props);
```
kyliepace: should there be an else statement to give a value to onBack?

PR #167, opened 2017-12-08
Allow users who choose both ID and DL to designate which will be Real ID

LeDeep:


client/containers/apply/real-id-container.jsx
```diff
@@ -14,22 +17,58 @@ const ConnectedForm = (props) => {
...
-  return (
-    <Form
-      onSubmit          = { onSubmit }
-      onBack            = { onBack }
-      onChange          = { props.onChange }
-      realID            = { props.realID }
-      cardType          = { props.cardType }
+  let content = [];
+
+  if(props.cardType.ID && props.cardType.DL) {
```
kyliepace: is creating separate forms for RealIdOneCard and RealIdTwoCard recommended over expanding on the <h4>Do you plan on using {card}</h4> variable logic? Asking to better understand what's preferred.

LeDeep: One option would be to assign the two distinct <h4> attributes to variables. And then show the one we want based on user selections. Might be cleaner, since we would only need one file. Let me make a commit.

logicgraphy:
It should also update the logic to card_options in database for realID. By default database will save create a record in card_options table for both the cards, if selected.

PR #168, opened 2017-12-11
Visual design, basic radio buttons integrated into the Voter Opt-out form

baccigalupi:
Hey, so I ran into some pretty big issues with the code base trying to get this in. I am going to copy in the important commit messages here, but on a bigger front the code for pre-registration made it impossible to integrate this small change with master (I was rebasing).

Any logic or abstraction should be focused on increasing the readability and changeability of the code. HTML is a very literal universe of strings and markup. React and Redux have a good pattern of keeping this presentation land of literals free of logic. When there is a ton of if/else statements constructing the text and the markup, the component is really hard to just glance at and understand. Worse when there is a ton of logic just changing the language in a template and other integrations become hard and therefore expensive.

I was able to make a clean base class for VoterOptOut form where all the differences in text could be injected as children, but as soon as I got that done, the logic in the next page was broken. We really need to revert all the commits in that feature set, and I am not sure if that is possible since it was merged instead of rebased.

Anyway will work on that problem tomorrow. Here are the other commit notes:

-------

## Integrates the RadioSelector into the Voter opt-out page

I was going to find a place that used the existing tappable buttons that
act like radios, but since this are already existed as radios, it seemed
like a simpler place to start.

This commit switches to using redux to store ui state as well as the
data going to and from the database.

I stopped using the connectForm module since there was a lot of
extra handlers for blur and focus. We might need to rethink the
connectForm to be more like a connectPage module. I am hoping a full
integration of all the selectable buttons will lead to some refactoring
and get to better abstractions. I think that on a per-page basis there
is a lot of duplication that we can build into one or more abstractions:
maybe one for ordinary page forms, one for read and continue pages and
one for expanding pages.

The other interesting change in this commit is that I switched from
saving the exact text shown on the page as the value of the radio button
to a symbolic answer. That means that when design or stakeholders change
the content of the page, we don't have to update all the data in the
database. We just update the presentation. Also, when we
internationalize, the logic built on recognizing default text will fall
apart.

Lots of changes to the feature specs around this. I haven't run the
server tests yet to see how the parsing might change the database ...
maybe another commit after this.

## Data parser for voter opt-in record and choices simpler and correct

The mapping of record values to ui values is really just one constant.
Separating those constants into two different functions means that there
is a high risk of introducing accidental bugs as the literals diverge.

Because I didn't want to introduce those bugs as I did this work, I
extracted out the constant map from the logic to go back and forth.

## Fixes tests that didn't break ? :( ?

So, the last commit changed the way the data was being passed to the
server, and it switched the way that the parser transformed that data,
but it didn't actually break any tests which made me worried. I wouldn't
expect the unit tests, based on mock data to break, but I did expect the
acceptance tests to fail.

This commit changes the tests, but does not address the lack of failure.
It extracts out the logic around transforming this particular data from
record to UI value and back again. It changes the mock data, etc.

I also noticed that the tests for the client and server parsing used
helper instead of just referring to literals or to the mock data. That
can get pretty unsafe since we don't know which thing is failing when we
have a breaking test. Both modules are capable of breaking. We will want
to change those tests to refer to the mock data or to be literals soon.


logicgraphy:
The work from this PR is integrated with PR##175 and rebased & merged  into master branch.

PR #169, opened 2017-12-11
youth-docs

kyliepace:
- redid previous youth-docs and youth-message PRs
- no longer store age in redux instead call getCurrentAge function in each container that needs it -
- wrote unit tests for ages 14, 15.5, and 17


client/containers/apply/date-of-birth-form-container.jsx
```diff
@@ -8,15 +8,12 @@ import connectForm            from '../../helpers/connect-form';
 import navigateOnSubmit       from '../../helpers/navigate-on-submit';
 import navigateOnBack         from '../../helpers/navigate-on-back';
 import * as dataPresent       from '../../helpers/data-present';
-import { getCurrentAge }       from '../../helpers/calculate-age';
```
baccigalupi: Good catch cleanup!

client/containers/apply/organ-donation-form-container.jsx
```diff
@@ -3,8 +3,9 @@
 import React from 'react';
 
 import { updateOrganDonation }      from "../../actions/index";
+import { getCurrentAge }            from '../../helpers/calculate-age';
 import HomeLink                     from '../../presentations/home-link.jsx';
-import NavigationButtons                from '../../presentations/navigation-buttons.jsx';
+import NavigationButtons            from '../../presentations/navigation-buttons.jsx';
```
baccigalupi: Ditto. Love that you are cleaning up these really small clutter things because clutter makes it hard to pay attention to the right places

client/containers/apply/organ-donation-form-container.jsx
```diff
@@ -14,7 +15,7 @@ import * as dataPresent             from '../../helpers/data-present';
 
 const ConnectedForm = (props) => {
   const continueDisabled = !dataPresent.organDonation(props.organDonation);
-  const onSubmit         = props.dateOfBirth.age >= 16 ? navigateOnSubmit('/voting-registration/introduction', props) : navigateOnSubmit('/summary', props);
+  const onSubmit         = props.age < 16  && props.age ?  navigateOnSubmit('/summary', props) : navigateOnSubmit('/voting-registration/introduction', props);
```
baccigalupi: This is the most important logic for the age boundary, and this is the section that I would separate out into the age related module and test like crazy. The tests would be for the day before a person turns 16; the day they turn 16 and then the day after to catch the full boundary logic in tests.

Embedded here means that the unit tests have for this logic have to render the page, and follow the next button in order to assert the age test is correct. Also, the main thing that changes is the path, the call to navigateOnSubmit is consistent in each branch.

client/containers/apply/organ-donation-form-container.jsx
```diff
@@ -42,7 +43,7 @@ const ConnectedForm = (props) => {
 function mapStateToProps(state) {
   return {
     organDonation: state.application.organDonation,
-    dateOfBirth: state.application.dateOfBirth
+    age: getCurrentAge(state.application.dateOfBirth)
```
baccigalupi: If the stuff above changes than we still just need the dob here.

client/containers/apply/what-do-you-want-to-do-today-container.jsx
```diff
@@ -12,7 +13,7 @@ import * as dataPresent       from '../../helpers/data-present';
 
 const ConnectedForm = (props) => {
   // if under 14 and selected "driverLicense", then instead of going to real-id page, go to page with message that user is too young and can only apply for ID
-  let onSubmit          =   props.age > 0 && props.age < 14 && props.cardType.DL === true ? navigateOnSubmit('/minor-DL-message', props) : navigateOnSubmit('/real-id', props);
+  let onSubmit          =   !(props.cardType.DL === true && props.age > 0 && props.age < 15.5) ?  navigateOnSubmit('/real-id', props) : props.age < 14 ? navigateOnSubmit('/minor-DL-message', props) : navigateOnSubmit('/youth-license-notification', props) ;
```
baccigalupi: Same idea with this logic which the important part to thoroughly test. It should be in a separate module (age one?). Separating it into a function means that we can give it a good name too such as `isTooYoungToApplyForLicense(age)`.

client/presentations/apply/youth-license.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+import SelectorCollection   from '../selector-collection.jsx';
+import NavigationButtons    from '../navigation-buttons.jsx';
+
+
+const Form = (props) => {
+
+  let notification = '';
+
+  if(props.age < 15) {
```
baccigalupi: This is more logic that I would want separated into a well named function and tested. I would also want to makes sure to use the same function if it is the same logic in the form.

test/client/helpers/calculate-age-test.js
```diff
@@ -5,32 +5,115 @@ import { getCurrentAge } from '../../../client/helpers/calculate-age';
...
-  describe('user is under 16', function(){
+  describe('age 16 boundaries', function(){
 
     it('user age is less than 16', function(){
       let year  = (new Date().getFullYear()) - 15;
       let month = '10';
       let day   = '15';
-      let age   = getCurrentAge(year, month, day);
+      let age   = getCurrentAge({year: year, month: month, day: day});
       assert(age < 16, 'age should be less than 16');
```
baccigalupi: This test is a bit funky since the test contains the logic we would ultimately like to document and separate as having business value ... but I think I already called out the functions that it would be good to see pulled out.

test/client/helpers/calculate-age-test.js
```diff
@@ -5,32 +5,115 @@ import { getCurrentAge } from '../../../client/helpers/calculate-age';
...
+      let age = getCurrentAge({
+        year: 2000,
+        month: 2,
+        day: 20
+      }, now);
+
+      assert(age >= 17, 'age should not be less than 17');
+    });
+  });
+
```
baccigalupi: I do really appreciate that there are so many more unit tests on this.

PR #170, opened 2017-12-11
 Update database to apply RealID selection to the appropriate card, w…

LeDeep:
…hether ID or DL


server/models/client-data-parser.js
```diff
@@ -253,15 +253,18 @@ function extractCardOptions(data) {
...
     })
-    if(data.realID.getRealID === 'Yes') {
-      cardOptions.push({
-        type:               'DL',
-        option_type:        'modification',
-        option_value:       'real-id'
-      })
-    }
   }
+
```
baccigalupi: How does this get associated with the DL card? Seems like you need to associate it with the right card_id at this point. But maybe I am not understanding the code. Can you explain?

server/db/migrations/20171211140452_updateCardIDTypeToInteger.js
```diff
@@ -0,0 +1,12 @@
+
+exports.up = function(knex, Promise) {
+  return knex.schema.alterTable('card_options', function(table) {
+    table.integer('card_id').alter();
+  });
+};
+
```
baccigalupi: Is this fixing a mistake in the type of the id as it was originally setup?

LeDeep: Yes. The card_id was originally a string, but it needs to be an integer.

server/models/client-data-parser.js
LeDeep: This happens in the server_data_parser.js, where it assigns the real ID designation to the selected card type, when card.id === card_id.

PR #171, opened 2017-12-12
Implements FAQ drawers for real ID feature page

logicgraphy:
- Implements FAQDrawer base class to hide/show the body/content.
- Adds actions and reducer to manage state of FAQ drawers
- Implements redux connect to dispatch action to toggle state of FAQ drawers in the redux.
- The name faq-drawer-ctrl is to be renamed after the existing faq-drawer is refactored and removed.


client/containers/apply/real-id-container.jsx
```diff
@@ -38,6 +40,18 @@ const ConnectedForm = (props) => {
     )
   };
```
baccigalupi: Looking at the whole length of the review, I think there is a simpler way to approach this with children. The pattern of putting together a page as a collection and rendering at the end works really well when there are a ton of parts that may or may not be rendered.

In the case of the FAQ drawer, the container is always the same, the content can be very complex markup containing other components.

My hope would be to see something that could be built like this:

```
<FAQDrawer {...props} title='What is a federally compliant card?'>
  <p>Some text</p>
  <a href='//somewhere-else' target='_blank'>and a link</a>
  <MaybeAComponent />
</FAQDrawer>
```
The drawer could know from state whether to show the children or not.

PR #172, opened 2017-12-12
selected reduced fee

kyliepace:
- changed default form view to have no buttons selected on both parts of the reduced fee form
- pulled logic out of presentations and all into the reduced fee container


client/containers/apply/reduced-fee-container.jsx
```diff
@@ -9,10 +9,12 @@ import navigateOnSubmit       from '../../helpers/navigate-on-submit';
 import navigateOnBack         from '../../helpers/navigate-on-back';
 import * as dataPresent       from '../../helpers/data-present';
 
+
 const ConnectedForm = (props) => {
   let onSubmit          =   navigateOnSubmit('/get-started', props);
   let onBack            =   navigateOnBack('/real-id', props);
   let continueDisabled  =   !dataPresent.reducedFee(props.reducedFee);
+  let DLText            =   props.cardType.ID && props.cardType.DL ? 'This only applies to your ID Card. You cannot get a free or reduced fee driver license.' : '';
```
logicgraphy: maybe we can change the logic to something like this:
```
const DLText = (props) => {
if (!(props.cardType.ID && props.cardType.DL)) {
  return null;
}
return <h5>This only applies to your ID Card. You cannot get a free or reduced fee driver license.</h5>
```
With this, we will not have black <h5> elements on the HTML.

logicgraphy:
I have made a small suggestion. But over all . its looks good and feel free to go ahead and merge this PR.

PR #173, opened 2017-12-12
Preregistration for youth from 16 to 18

teresavshen:


test/client/helpers/calculate-age-test.js
```diff
@@ -3,34 +3,113 @@
 import assert from 'assert';
 import { getCurrentAge } from '../../../client/helpers/calculate-age';
 
-describe('calculate user age', function() {
+describe('calculate user age', function () {
 
-  describe('user is under 16', function(){
+  describe('age 15 boundaries', function () {
```
baccigalupi: These tests setup boundary conditions inside the tests, and then test them. What we really want in a function that sets up the boundary conditions and we thoroughly test that boundary condition.

I think that because one of the first things we learn in programming is conditionals it is easy to miss that most bugs come from incorrect conditionals. Extracting that logic and attaching a good name to it helps. Even more helpful is exhaustive testing so that when product comes along and changes it (or contests its logic) you can stay exactly how the code is working by the day or hour.

baccigalupi:
If you can fix these boundary date extraction things by the time we get to our 1pm meeting, that would be great. Otherwise we can do it in the 1pm meeting and then do a rebase.

PR #193, opened 2017-12-19
Tests and re-arranges logic and presentation in youth notification page:

baccigalupi:
I smooshed this into one commit, which makes it harder to see how I changed the individual files, since I also moved them. Let me know if you want a walk through.

----------

Moves header logic and error notification into page presentation and out
of the container.

Writes tests for the page presentation to ensure that partials are
rendered according to expected logic.

Removes data change from components and moves it into the reducer for
card update. Logic in the original reducer assumed true and false as
strings, but it looks like they are true and false values instead.

Changes where the error and other information shows up on the page.

Moves youth notification to `intro` directory. Also fixes some
whitespace.


client/containers/intro/youth-license-notification-page.jsx
```diff
@@ -0,0 +1,52 @@
...
+import {
+  updateCardType
+} from '../../actions/index';
+
+import onInputChange                  from '../../helpers/on-input-change';
+import onFormSubmit                   from '../../helpers/on-form-submit';
+import navigateOnSubmit               from '../../helpers/navigate-on-submit';
+import navigateOnBack                 from '../../helpers/navigate-on-back';
+import * as dataPresent               from '../../helpers/data-present';
+import { getCurrentAge, ageChecks }   from '../../helpers/calculate-age';
```
kyliepace: can delete getCurrentAge from this container

baccigalupi: Yes!

PR #204, opened 2017-12-21
choose card action (renew or apply) before choosing card type(s)

kyliepace:
- what-do-you-want-to-do-today presentation changed to radio button form letting user change cardAction to either "new" or "renew"
- inserts /apply/select-id-dl page letting user select ID or DL card type
- if renewing, choose-card-type presentation has a radio button form and changes the value of props.cardType.renew
- if getting a new card, choose-card-type presentation has a checkbox form and changes the value of props.cardType.new array
- updates all instances of cardType.DL === true or cardType.ID === true logic to instead use functions defined in helpers/data/card-type.js
- if user is renewing card, /apply/current-card-information is inserted between choose card and real id pages
- current card info saved to new db table
- tests added


kyliepace:
waiting on feedback on feature testing PR before changing the feature tests in this branch, since merging the feature testing PR will change 99% of the feature tests in this branch

PR #210, opened 2017-12-28
protect against cross-site scripting with a CSRF Token

kyliepace:
- added [csurf](https://github.com/expressjs/csurf) library and cookie-parser
- instructed express app to require CSRF Token issued by csurf library 
- send cookie and token cookie to user; token cookie has sameSite, httpOnly, and secure flags
- use Fetch API's credentials parameter to send cookies back to server when the createApplication function is called, so the server can compare and ensure that this is the intended user making the request
- [Google doc explaining choices ](https://docs.google.com/document/d/1dXgf7DJY8BWJCWDxM8ypHJVBDp2UV_rZdrBuJUKMxCI/edit)


server.js
```diff
@@ -1,29 +1,39 @@
...
 server.use(bodyParser.json());
+server.use(cookieParser());
+server.use(csrf({ cookie: true, value: (req) => (req.cookies.csrfToken) }));
 server.use(logging());
 server.use(helmet());
 
 server.port = env.port;
 server.environment  = env.env;
 
+server.use(function (req, res, next) {
```
baccigalupi: This too would be better in another place, controller/middleware area or in config.

Also, wondering about putting it before the static content. Is that necessary?

server.js
```diff
@@ -1,29 +1,39 @@
...
+const cookieParser  = require('cookie-parser');
+
+const layout        = fs.readFileSync(path.resolve(__dirname, 'server/templates/layout.html')).toString();
+let   server        = express();
 
 passport.use(jwtStrategy);
 server.use(passport.initialize());
 server.use(bodyParser.json());
+server.use(cookieParser());
+server.use(csrf({ cookie: true, value: (req) => (req.cookies.csrfToken) }));
```
baccigalupi: It would be great to see the body of this extracted into the config area. For this file, I want to keep the noise and detail down so we can see the exact nature of how the app is setup with well named configurations. See the `logging()` config below for an example.

PR #218, opened 2018-01-02
Edit name from summary

kyliepace:
- https://www.pivotaltracker.com/story/show/153711680
- created two new components, page-summary-link, and summary-item, to make the new design of info on the summary page consistent
- each page-summary-link has a link to the section's form url
- extracted summary presentation out of the container, following example of other pages in app
- when the user reaches the summary, the state.ui.edit variable is toggled. then, when the user clicks to edit info on a form, clicking the "next" button on the form will take them back to the summary instead of taking them to the next form in the app
- deleted some unused, old versions of containers


kyliepace:
idea - can pass string to the names page container through [link location object](https://stackoverflow.com/questions/45598854/passing-props-through-react-router-v4-link)

kyliepace:
ready to be reviewed again

PR #224, opened 2018-01-04
Org sections

baccigalupi:
This isn't really ready to go. There is some kind of problem with phantom that I haven't figured out yet.

@kyliepace Wanted you to check this out. Saw in #developers that you found a simpler solution. While I was breaking things up into sections I started to see a missing data object, from there I started to see ways to extract navigation and centralize it around that data. Then I found a good place to insert edit mode changes.

See what you think and if it helps or if your solution is better.


client/helpers/data/page.js
```diff
@@ -0,0 +1,31 @@
...
+export const pageFor = (name) => {
+  return all[name];
+};
+
+export const pathForPage = (name) => {
+  return pageFor(name).path;
+};
+
+export const nextPath = (name, props) => {
+  if (props.editMode === true) { return pageFor('summary').path; }
```
kyliepace: I like this solution! It's also compatible with what I did for my second attempt at this problem, which was to get rid of setting props.editMode through ComponentWillMount() by sending a prop to the container via the Link location object. I think this nextPath function could just check for that value instead.

baccigalupi: Cool, cool. I don't suppose you have any idea what is breaking phantom?

kyliepace: No, looking at it now.

kyliepace: I had an issue like this once from using Object.entries so I thought it might be Object.keys that is causing the problem, but that's been used already in the validations branch and didn't cause any issues.

baccigalupi: Yeah, I did a search for Object.keys and it is already used

kyliepace: seems like the test is breaking at the waitForSelector('.legal-name-form') line in the navigation-steps

baccigalupi:
@kyliepace It was a missing route, the default one :( I didn't think you could get there because of the server, so I deleted it, and everything broke.

PR #225, opened 2018-01-04
Voting registration section render the header from state

teresavshen:
- https://www.pivotaltracker.com/n/projects/2051881/stories/153866870
- Voting registration section render the header from state
- Refactored voting registration pages


client/containers/voter-registration/contact-methods-form-container.jsx
```diff
@@ -1,89 +1,63 @@
...
+  let onSubmit            = handlers.navigateOnSubmit('/voting-registration/confirmation', props);
+  let onBack              = handlers.navigateOnBack(props);
 
   if (isPreregistering(props.dateOfBirth)) {
-    return <PreRegContactChoice
-      pageTitle={formPageTitle}
-      sectionName={formSectionName}
-      onChange={props.onChange}
-      selectedValue={props.contactMethods.shouldContact}
+    return <PreRegContactMethodsPage
```
kyliepace: should this be consistent with how Presentation and isPreregistering is handled in the other containers?

PR #229, opened 2018-01-05
Correct or update page

kyliepace:
- https://www.pivotaltracker.com/story/show/154001641
- add radio option on wdywtdt page to let user change (correct or update) card
- if selected, user gets taken to /updates-and-corrections page to specify correcting or updating and to select which section of the form they would like to change
- data saved to redux and database


LeDeep:
Looks good! I think the one change would be the navigation from the 'select-id-dl' page. Looking at the scenario, after making the card selection and clicking next, the user should go to the 'current-card-information' page and then from there to the 'updates-and-correction' page. Currently it's reversed.

PR #249, opened 2018-01-09
Validate card type

logicgraphy:
- Implements validation for choose card type page.
- Refactors choose card type checkbox selector page to warp within a checkbox selector collection.
- Adds validation error messages to checkbox selector collection component.
- Fixes tests.


client/containers/intro/choose-card-page.jsx
```diff
@@ -34,28 +34,46 @@ const addressForProps = (props) => {
...
-  let continueDisabled  =   !canContinue(props)
 
   return (
     <Presentation
       {...props}
       onSubmit          = { onSubmit }
       onBack            = { onBack }
-      continueDisabled  = { continueDisabled }
+      validations       = { validations }
+      focus             = { focus }
```
kyliepace: I think the focus isn't working right on the radio buttons, but I think that's because this focus variable on line 64 doesn't point to any actions and the form needs to call props.onFocus and props.onFocusClearValidation

PR #262, opened 2018-01-11
Validations for youth dl

teresavshen:
- Validations for Youth DL message page


client/containers/intro/youth-license-notification-page.jsx
```diff
@@ -1,30 +1,39 @@
...
-import { getDL }            from '../../helpers/data/card-type';
-import { ageChecks }        from '../../helpers/calculate-age';
+import Presentation               from '../../presentations/intro/youth-license-notification-page.jsx';
+import { updateCardType }         from '../../actions/index';
+import { getDL }                  from '../../helpers/data/card-type';
+import { ageChecks }              from '../../helpers/calculate-age';
 
 const Page = (props) => {
+  let validations         = new YouthDLValidator(props.cardType.youthIDInstead, props.validations);
   const continueDisabled  = ageChecks.Under15(props.dateOfBirth) ? props.cardType.youthIDInstead !== 'Yes' : false;
```
kyliepace: now that validations are being used the continueDisabled can be deleted, except I know this page is special because of how the next button is supposed to disappear if the user is under 15.5 and clicks No to getting an ID instead. Is that being implemented through the continueDisabled function here?

teresavshen: yes, continuedisabled is being used because of special circumstance (under 15.5 and no to ID is selected). it's separate from validation.

PR #286, opened 2018-01-17
secondary help labels

kyliepace:
- made secondary MM, DD, YYYY labels not bold
- added unit testing for DateInput and NumberInput components
- individualized the error passed to NumberInput so that the individual input on Date entries gets a red outline if not valid, rather than all 3 inputs getting red outline if only one is [invalid](https://www.pivotaltracker.com/story/show/154379398)


test/client/presentations/number-input-test.js
```diff
@@ -0,0 +1,64 @@
...
+  it('passes a string to a bold label above the number inputs', function() {
+    let component = render(
+      <NumberInput {...props} />
+    );
+    assert.ok(component.find('label:not(".normal")').length, 'bold label not found');
+    assert.equal(component.find('label.normal').length, 0, 'secondary labels found');
+  });
+
+  describe('error states', function() {
+    it('when there is an error, it will have additional text and the bold label and the input will get the error class and there will be an error icon', function() {
```
LeDeep: I think each assertion should have its own `it` block. It will read a little better, and make the tests easier to follow. You could break this long statement into three `it` blocks. Same for the date-input-test.js file.

kyliepace: good point! I have made those changes and pushed to this branch

PR #293, opened 2018-01-17
Validations for Social Security

teresavshen:
- https://www.pivotaltracker.com/n/projects/2051881/stories/154192085


client/helpers/data/next-path.js
```diff
@@ -66,6 +66,18 @@ export const realID = (props) => {
   return key;
 };
 
+export const ssn = (props) => {
+  let key = 'myBasics';
```
LeDeep: You don't need `mybasics` here. You set the default `key` to `licenseHistory` and get rid of the `else` statement.

PR #297, opened 2018-01-18
Validate address

logicgraphy:
- Adds validations to address template based on the type. 
- Moves mailing and home under single address state/redux object.
- Modifies client and server data parsers to work with single address object.
- Fixes existing tests and adds unit tests for address reducer and validation rules.


client/helpers/validations/address-rules.js
```diff
@@ -0,0 +1,95 @@
...
+import { hasValue }       from '../data/validations';
+import { hasOnlyNumbers } from '../data/validations';
+import {
+  englishValidatorGenerator,
+  englishValidator
+} from './english-validator';
+
+const homeAddressSameAsMailing = (props) => {
+  let value = props.homeAddressSameAsMailing;
+  let errors = englishValidator(value);
```
LeDeep: This is just a yes/no selector, so you don't really need the `englishValidator`. `errors` could just be an empty array.

PR #331, opened 2018-01-25
Cardtype refactor

kyliepace:
- added DL and ID objects that track whether the user is getting that card and what action is happening with that card
- see update-card-type reducer and associated files

state looks like:
`{
IDDL: [],
cardAction: '',
youthIDInstead: '',
DL: { isApplying: false, action: ''},
ID: { isApplying: false, action: ''}
}
`
- I kept the IDDL array for ease of use and consistency with checkbox collection component. Is that a large compromise for a single component?     maybe?

- the cardAction key is kept only for validations for the cardAction page, and isn't saved to the server. The server gets its cardAction info from the card objects' actions.
- since I kept the IDDL array, the card object isApplying key is redundant and is not actually used currently

- I added space to handle what I predict will be a future user story, namely having two cards with two different actions (eg renewing an ID, replacing a DL) and editing the action on only one of them from the summary screen. the reducer currently wouldn't work for this possible future scenario, but I think it is built so that it could.


kyliepace:
- worked with Teresa to think better about how the youthID notification page needs to change the cardType in redux. Added logic to the youthID reducer so that a user who is applying for two new cards doesn't have either lost by redux when they select "No" to getting an ID instead.
- added to and refactored the reducer unit tests
- multiple card distinction is made by the youth notification page using a helper function and passing result to the presentation

PR #357, opened 2018-01-31
extract all logic refactoring: my history section

kyliepace:
- extracted remaining logics from my history presentations and containers into helpers and added testing to all helper functions. 

- changed the prop on the dateInput from description to title to avoid overlap with child number input's prop names

- moved the conditions that determine whether a component will render null to the main presentation view to make it easier to quickly look at the presentation and understand what's happening. is this a bad idea?


client/helpers/data/veteran.js
```diff
@@ -6,39 +6,65 @@ import {
   isGettingNew,
   isCardActionSelected
  }    from './card-actions'
+ import { getDL }             from './card-type';
 
 export const mustChooseBenefits = (props) => {
```
logicgraphy: I think this function should be named like 'mushChooseIsVeteran'. Benefits selection is a sub-page under veterans. Will be nice to have some consistency is the names.

kyliepace: how about just "isVeteran" like the "isPreviouslyDesignated" name on line 40?

PR #367, opened 2018-02-02
extract logic from get-started section

kyliepace:
- created functions for determining which string should be shown, for example function getStringByAction in helpers/data/card-actions.js. A presentation like the choose-card-page in presentations/get-started/ passes all the possible strings to that function and gets the correct one back. Is this too complicated a solution?

- wrote some custom logic functions for the get-started intro page presentations

- thinking ahead, it seems like we might reach a point where our logic needs to distinguish between the card type and card action of the current flow, and those of the entire application. when I was really in the weeds yesterday afternoon I created a file and wrote some placeholder-type logic functions in helpers/data/summary.js that would consider the entire application, not just what flow the user is going through (e.g. adding a renewal ID card when they have also already chosen to get a new DL). we don't need them at this time and this might not be how we want to handle it but I left them in, in case


client/helpers/data/card-actions.js
```diff
@@ -12,10 +13,56 @@ export const isReplacingCard = (props) => {
...
+
+export const hasActionIsCorrecting = (props) => {
+  return isChangingCard(props) && isCorrecting(props);
+};
+
+export const hasActionIsUpdating = (props) => {
+  return isChangingCard(props) && isUpdating(props);
+};
+
+export const getStringByAction = (props, newString, renew, replace, change, update, correct) => {
```
logicgraphy: This is pretty handy to generate context. Will be nice to reduce some logic complexity (https://en.wikipedia.org/wiki/Cyclomatic_complexity) here.
Thinking, might be better move this into redux state? What do you think?

client/helpers/data/card-type.js
```diff
@@ -1,55 +1,87 @@
...
-export const IDorDL = (props) => {
-  return hasMultipleCards(props) ? 'both' : getID(props) ? 'ID' : getDL(props) ? 'DL' : 'none';
+export const needsEndorsement = (props) => {
+  return props.licenseType.needEndorsement === 'Yes';
 };
 
-export const IDOnly = (props) => {
-  return getID(props) && !getDL(props);
-};
+export const getCorrectString = (props, DLString, IDString, bothString) => {
```
logicgraphy: This state that can also be maintained within redux.

client/helpers/data/card-actions.js
logicgraphy: On a second thought, it will make internationalization difficult if we move strings into redux. The current approach looks good. If we can, will be nice to simplify the string building logic a little bit. Please see if this can be simplified, otherwise feel free to merge this into master.

PR #398, opened 2018-02-14
added renewal button to add DL flow

kyliepace:
- holding off on redux state solution because for now there's no button starting off the new add DL flow so the redux state would have to be triggered from the container loading, and I don't think that's the solution we want. Other thoughts though?


kyliepace:
there's another PR that improves the codeclimate score on the next-path and pages files

PR #429, opened 2018-02-24
Oauth2 Authentication with id.me

logicgraphy:
- Implements oauth2 passport strategy for authenticating with id.me
- Adds migration scripts for authentication tables/columns, saves user data into DB.
- Adds link on client side to initiate login/signup onto id.me
- Implements express-session to store user info into session, more work on this in future PR.
- Removes JWT related strategy, env variables and tests.


client/containers/get-started/id-me-page.jsx
```diff
@@ -7,16 +7,16 @@ import handlers           from '../../helpers/handlers';
 import Presentation       from "../../presentations/get-started/id-me-page.jsx";
 
 const Page = (props) => {
-  let onBack = handlers.navigateOnBack(props);
+
+  let onBack              = handlers.navigateOnBack(props);
+  const authorizeWithIDME = () => window.location = '/auth/new';
```
baccigalupi: Cool!

server.js
```diff
@@ -1,23 +1,37 @@
...
+const oauth2Strategy        = require('./server/config/oauth2-strategy').strategy;
+const sessionOptions        = require('./server/config/session');
+const logging               = require('./server/config/logging');
+const csrf                  = require('./server/config/csrf');
+const api                   = require('./server/api');
+let startingIndex           = process.env.APP_ENV === 'development' || process.env.APP_ENV === 'test' ? 'index.dev.html' : 'index.html';
+const layout                = fs.readFileSync(path.resolve(__dirname, 'public/'+startingIndex)).toString();
+let   server                = express();
+
+const requireLogin = (req,res,next) => {
```
baccigalupi: Can you move this into a separate file in `controllers` or `middleware` ?

server.js
```diff
@@ -41,10 +55,14 @@ server.get('/add*', (req, res) => {
   res.send(layout);
 });
```
baccigalupi: These should be in a combination of routes and controllers. The server file should just have configuration details, with the logic extracted. The reason is so that the security and customization of the server are easily read (and modified).

Same is true for the routes separate from the controller. If there is one big routes file, it is easy to see all the paths in the app and figure out which ones correspond to which module/controllers.

server/config/env.js
```diff
@@ -10,7 +10,7 @@ if (env === 'development' || env === 'test') {
 const port   = process.env.PORT || getPort(env);
 
 function getPort(env) {
-  return env === 'development' ? 3000 : 3333;
+  return env === 'development' ? 80 : 3333;
```
baccigalupi: Is this necessary? It messes with dev machines having to hand over 80 to a particular app.

server/config/oauth2-strategy.js
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const authorizeApiUsage = function (accessToken, refreshToken, profile, done) {
+  findOrSaveUser(profile)
+    .then((user) => {
+      return done(null, user);
+    });
+};
+
+module.exports = {
```
baccigalupi: The indentation looks weird here.

server/db/migrations/20180222135627_add_authentications_table.js
```diff
@@ -0,0 +1,24 @@
+
+exports.up = function(db, Promise) {
+  return Promise.all([
+    db.schema.table('applications', function(table) {
+      table.string('auth_uuid');
```
baccigalupi: For sure we want to use the `authentications.id` as the foreign key since it is an integer and faster to search and index.

test/server/config/oauth2-strategy-test.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+const assert            = require('assert');
+const Oauth2Strategy    = require('../../../server/config/oauth2-strategy');
+
+let authorizer          = Oauth2Strategy.authorizeApiUsage;
+let Strategy            = Oauth2Strategy.Strategy;
+
+describe('Oauth2Strategy authorizeApiUsage', function() {
```
baccigalupi: I'm confused by this test. It says that it finds or saves, but no db pre-work is done. Also those should be separate tests to make sure they both work.

test/server/config/oauth2-strategy-test.js
```diff
@@ -0,0 +1,85 @@
...
+    authorizer('aToken', 'rToken', {uuid: 'A123'}, function(err, user) {
+      assert(user[0].uuid);
+      done();
+    });
+  });
+});
+
+describe('Oauth2Strategy userProfile', function () {
+  describe('loading profile', function () {
+      var strategy =  new Strategy({
```
baccigalupi: Indentations in this file look bad. Might be time to check IDE's and make sure they are set to two spaces and not tabs or four spaces.

test/server/config/oauth2-strategy-test.js
```diff
@@ -0,0 +1,85 @@
...
+          if (url !== 'https://api.idmelabs.com/api/public/v3/attributes.json') {
+              callback(new Error('wrong url argument'));
+              return;
+          }
+
+          if (accessToken !== 'token') {
+              callback(new Error('wrong token argument'));
+              return;
+          }
+
```
baccigalupi: I would typically put this is a json file in support or fixtures or some other test helper area.

test/server/db/models/save-user-test.js
```diff
@@ -0,0 +1,46 @@
...
+      findOrSaveUser(data)
+        .then((record) => {
+          assert.equal(record[0].uuid, data.uuid);
+          assert.equal(record[0].full_name, data.full_name);
+          assert.equal(record[0].email, data.email);
+        })
+        .then(done)
+        .catch(done);
+    });
+
```
baccigalupi: These tests look the same, and the `beforeEach` blows away the database with each run. That is really good. We don't want any interdependency between tests. But I think you need a beforeEach block that saves data before running the second test.

server/config/env.js
logicgraphy: This was required because the callback from id.me corresponds to lvh.me and that runs on port 80.

PR #432, opened 2018-02-26
Real id summary

kyliepace:
- another stab at updating how real ID info is saved in redux
- fixes bug described in story #155449035
- realID follows the cardType pattern where both the IDApp and the DLApp objects have their own separate realID string, and the application has a base-level realID string to fall back to in the case of a user getting two new cards in the same flow

`state: {
application: {
realID: '',
IDApp: {
realID: ''
},
DLApp: {
realID: ''
}
}
}`

- because the addApp state is set from the url pathname, I created two new routes so that when the user clicks to edit the realID info from the summary, the user is taken to one of the new routes and the addApp state updates correctly. After talking to Randeep last week I am more and more in favor of changing this so that clicking the edit button is what sets the addApp state, but I didn't make that change in this story.


kyliepace:
https://www.pivotaltracker.com/story/show/155449035

PR #440, opened 2018-02-27
add california-residency page to CDL flow

kyliepace:
#155155236

- adds residency object to CDL redux state
- adds container and presentation for new page
- reuses existing address presentations and validations


kyliepace:
https://www.pivotaltracker.com/story/show/155155236

PR #442, opened 2018-02-27
Just some small fixes on a current card info page

baccigalupi:
The design review had this page header noted to be changed, but it turns out it was handled by the translation process.

So I just swapped in the Translate tag on this page.


baccigalupi:
Oops, I just rebased this in accidentally. Sorry!

PR #458, opened 2018-03-05
CDL Social Security

teresavshen:
- https://www.pivotaltracker.com/n/projects/2051881/stories/155155242


client/containers/cdl/my-basics/cdl-social-security.jsx
```diff
@@ -0,0 +1,35 @@
...
+import handlers                      from '../../../helpers/handlers';
+import { SSNValidator }              from '../../../helpers/validations';
+import { updateCdlSocialSecurity }   from '../../../actions/index.js';
+import Presentation                  from '../../../presentations/cdl/ssn/social-security-page.jsx';
+
+const Page = (props) => {
+  let locale            =   props.locale;
+  let validations       =   new SSNValidator(Object.assign(props.socialSecurity, {locale}), props.validations);
+  let onSubmit          =   handlers.navigateOrShowErrors('cdlSocialSecurity', props, validations);
+  let onBack            =   handlers.navigateOnBack(props);
```
kyliepace: pass validations as the 2nd argument to the navigateOnBack function so that when the user edits this page from the summary, they can't click "back" to return to the summary if they have introduced errors onto the SSN page

PR #463, opened 2018-03-05
Cdl self-cert

kyliepace:
- created new page for CDL user to self-certify Interstate or Intrastate CDL
- did not attach to mergePropsGenerator but in the future might be the way to go to implement forthcoming user story about clicking next to certify agreement?
https://www.pivotaltracker.com/story/show/155374549
- reusing SelectionValidation form validator


logicgraphy:
Looks great! Will be nice to add a className='translation-missing' where it makes sense.

PR #473, opened 2018-03-07
Creates async action for getting translations after language choice

baccigalupi:
This series of commits handles just the aysnc actions for getting the data from the server and responding. It's pattern is based off the existing API work, and it extracts some actions that should probably be refactored back into the existing api actions.

Note: while this creates the async action and the onSubmit handler, it does not yet integrate them into page submission.

After this, I am going looking to hand this back to the team. Will write some notes on the ticket and we can talk in sprint planning about who can take this on next.


baccigalupi:
Seems like the test coverage reporting from code climate is weird. Test coverage for sure went up or stayed the same.

PR #492, opened 2018-03-12
added edit navigation logic for changes made to card action from the summary

kyliepace:
- clicking the "add" or "edit" button from the summary triggers the onFlowChange handler through the pageSummaryLink component (see files client/helpers/handlers/onFlowChange.js and client/presentations/page-summary-link.jsx), which sets the redux state.ui.flow. Previously the state was set every time the url changed from /id-and-license/ to /add/ or /edit/. 

- all navigation logic, for adding, editing, and normal flow, is handled in the pages.js and next-path.js files for each section. there is no separate add or edit folder. see file client/helpers/navigation/id-dl/get-started/next-path.js for an example of how the helper functions consider props.flow

- there are a lot of intermediate commits of different paths I took for this story that didn't make it to the final one. squash before merging?


baccigalupi:
I think this is too big a refactor for me to understand. Do you want to get on a screen sharing and take me through it?

PR #497, opened 2018-03-12
Implements translator tag with lookup fail over.

logicgraphy:
This PR is a part of the bigger work that is required for implementing internationalization for voting section. In this PR only a part of Citizenship page is implemented as a solution validation check point. Ones the solution is validated by team members, other pages will follow the same modifications.

- Connects the getTransaltion end point with the app.
- Stores en as the default translation in redux.
- Saves selected translation json in redux after fetching from server.
- Implements translation look up in selected language, if not found, then looks into default.
- Returns a blank string when no translation found for the given contentKey.
- Implements unit tests and presentational tests.


client/actions/index.js
```diff
@@ -47,6 +47,7 @@ export const TYPES = {
   UPDATE_VETERANS_SERVICE:                'UPDATE_VETERANS_SERVICE',
   UPDATE_CITIZEN_STATUS:                  'UPDATE_CITIZEN_STATUS',
   UPDATE_BALLOT_BY_MAIL:                  'UPDATE_BALLOT_BY_MAIL',
+  UPDATE_BALLOT_LANGUAGE:                 'UPDATE_BALLOT_LANGUAGE',
```
baccigalupi: Wondering about this change. Is this a missing action from the work we did before on automatically defaulting to the application language when that is chosen? On is this the way we are indicating the translation locale?

If it is about locale, I would change the name of the action so that it is more general. For this story we are changing the voter translations (just to keep the scope small and doable), but the next stories will fix the rest of the application.

client/containers/voter-registration/ballot-language-form-container.jsx
```diff
@@ -3,7 +3,7 @@
 import React                      from 'react';
 import connectForm                from '../../helpers/connect-form';
 
-import { updateLanguage }         from '../../actions/index';
+import { updateBallotLanguage }   from '../../actions/index';
```
baccigalupi: Oh, yeah I get it now. Ignore my previous question.

client/containers/voter-registration/ballot-language-form-container.jsx
```diff
@@ -28,13 +28,13 @@ const Page = (props) => {
...
+    ballotLanguage  : state.application.voting.ballotLanguage,
     dateOfBirth     : state.application.basics.dateOfBirth,
     focused         : state.ui.focus,
     validations     : state.ui.validations,
     locale          : state.ui.locale
   };
 };
 
-export default connectForm(mapStateToProps, updateLanguage, Page);
+export default connectForm(mapStateToProps, updateBallotLanguage, Page);
```
baccigalupi: Just to be clear about the story, before it gets to acceptance though, the ballot language shouldn't change the application language they have already chosen.

client/helpers/data/translator.js
```diff
@@ -0,0 +1,34 @@
...
+
+export const keyLookup = (contentKey, translation) => {
+  try {
+    let contentKeys = contentKey.split('.');
+    let tempObject  = translation;
+
+    for (var i in contentKeys){
+      tempObject = tempObject[contentKeys[i]]
+    }
+    return tempObject ? tempObject : '';
```
baccigalupi: Oh this is a good solution. I was thinking about using the null pattern instead of err catching. In that case you would return an empty object if not found `{}`. There would never be an error, but at the end you can test for truthiness.

This way is really good and easy to understand too.

client/helpers/on-submit-dispatches.js
```diff
@@ -43,37 +42,24 @@ export const saveApplication = (stateProps, dispatch, ownProps) => {
...
     if (choiceNotMade) {
-      dispatch(updateLanguage('appLanguage', 'en'));
+      dispatch(updateLanguage('language', 'en'));
     }
 
-    if (choiceNotMade || choiceIsEngish) {
-      return ownProps.history.push(
-        nextPath('chooseLanguage', stateProps)
-      );
+    if (choiceNotMade || choiceIsEnglish) {
```
baccigalupi: If there is a way to get rid of this empty if block that would be good. The conditionals are really easy to understand as they are named and this is written. So you would want to maintain that. Maybe you could make a third attribute `nonEnglishChoice` that is built from these others?

client/i18n/translator-tag.jsx
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
```
baccigalupi: Looks good. Maybe `contentKey` should be named with `path` instead. Like `contentPath` or `translationPath`.

client/presentations/get-started/choose-application-page.jsx
```diff
@@ -8,16 +8,27 @@ import RadioCollection          from '../radio-selector-collection.jsx';
...
+  let className = `choose-application-form ${hideMain(props)}`;
   return (
     <Page
       {...props}
       sectionKey='intro'
     >
-      <form onSubmit = { props.onSubmit } className='choose-application-form'>
+
+      <div className={props.server.apiStatus}/>
+
```
baccigalupi: I am wondering if this error message merits its own presentation now that it shows up in two places. Also, the logic here should move out of the markup. This looks to me like a guard clause in a separate component.

client/reducers/server/update-translations.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
```
baccigalupi: This looks good!

server/routes.js
```diff
@@ -9,7 +9,7 @@ const routes = (passport) => {
 
   router.get( '/api/application/:id',   controllers.getApplication);
   router.post('/api/application',       controllers.postApplication);
-  router.post('/api/translation/:code', controllers.getTranslation);
+  router.get('/api/translation/:code',  controllers.getTranslation);
```
baccigalupi: Good catch!

test/client/helpers/data/translator-test.js
```diff
@@ -0,0 +1,64 @@
...
+      assert.equal(keyLookup(contentKey, translation), '');
+      done();
+    });
+
+    it('empty string when an exception happens', (done) => {
+      let contentKey  = 'votingRegistration1.citizenshipPage1.pagePrompt1';
+      assert.equal(keyLookup(contentKey, translation), '');
+      done();
+    });
+
```
baccigalupi: It would be good to see a test case where the english exists, but the translations doesn't. That is going to be the case as we are waiting for translations and adding them in.

test/client/helpers/data/translator-test.js
```diff
@@ -0,0 +1,64 @@
...
+    it('returns translated value from selected language, if its available', (done) => {
+      let contentKey  = 'votingRegistration.eligibilityPage.convictionStatement';
+      assert.equal(translateThis(contentKey, props).__html, 'No estoy actualmente en una prisión estatal o federal o en libertad condicional por haber sido condenado de un delito mayor');
+      done();
+    });
+
+    it('returns translated value from default language, if its not available in selected', (done) => {
+      let contentKey  = 'applicationPreparationPage.prompt';
+      assert.equal(translateThis(contentKey, props).__html, 'Appointment preparation');
```
baccigalupi: Oh, here it is. Late to the party again.

baccigalupi:
Oh, but it looks like the tests aren't passing. So, looks like some changes are needed to make that happen.

PR #506, opened 2018-03-13
Add ballot by mail and contact methods to cdl voting registration flow

LeDeep:


client/helpers/navigation/cdl/voter-registration/pages.js
```diff
@@ -50,8 +50,26 @@ const cdlVoterRegistration = [
...
     path: '/voting-registration/language',
-    next: ''
-  }
+    next: 'cdlBallotByMail'
+  },
+  {
+    key: 'cdlBallotByMail',
+    description: 'Ballot by mail',
+    path: '/voting-registration/vote-by-mail',
+    next: 'cdlContactMethods'
```
kyliepace: this needs a helper function nextOrCDLSummary to determine if clicking next should take the user to the cdlContactMethods page or back to the cdlSummary page (if that's how this should behave). The path also could use the applyOrEditCDLPath to determine the route, but it won't break if it doesn't use it.

PR #515, opened 2018-03-16
Replace convertToHtml with <Translate> tag for consistency

LeDeep:


logicgraphy:
This is good. Helps with the translation lookup/refactoring I am working on.

PR #518, opened 2018-03-19
Cdl save application to DB

kyliepace:
- reused IDDL database structure except for adding military_waiver column to veteran table. made the address and the cardReplacement redux names consistent across both apps

- server-to-client parser looks to see if the card type is CDL and if it is, it gets the db data into the CDL data structure

- added more server tests


logicgraphy:
That was a heavy lift. Great work! If you can give team a walk through the over all approach and particularly with card options part, that will be great. Also, there are some acceptance tests that are commented; it will be nice to fix them.

PR #528, opened 2018-03-21
minify and gzip files for heroku and production environments

kyliepace:
- [demo at heroku-qa](http://dlite-web-dev-qa.herokuapp.com/apply/choose-application) app.js file should be ~250kb instead of >1mb.

- uses [expressStaticGzip library](https://github.com/tkoenig89/express-static-gzip)


logicgraphy:
Looks good! We do not want to check in the binary files, it should be added for git to ignore.

PR #529, opened 2018-03-21
CDL sign-in redirects to CDL legal name page

kyliepace:
- ID.me container sets document.cookie equal to either CDL or id-and-license, looking first for the redux ui.chooseApp value and if it's not there (like if we're in dev mode) then looking at the url

- upon successful sign-in the user is redirected to /apply/logged-in
- logged-in container reads document.cookie and pushes user to either /apply/cdl/true-name or to /apply/id-and-license/legal-name


kyliepace:
the cookie that saves whether the user has taken cdl or iddl path doesn't work on localhost when redirected to lvh.me

PR #533, opened 2018-03-22
Restyling selectors and faq drawers

baccigalupi:
Changes the selector colors and styling to match the new designs. Also changes the FAQ drawer to almost match the design, but adds border (just because I am opinionated).


baccigalupi:
Oh also, this commit adds fixes for gitignore with gzips

PR #534, opened 2018-03-22
CDL sign-in saves app choice and redirects to correct page

kyliepace:
- idme page saves appName cookie
- if in development mode, server redirects to localhost://apply/logged-in after successful signin so appName cookie is available
- logged-in page pushes to either /apply/cdl/true-name or to /apply/id-and-license/legal-name depending on the value of the appName cookie
- because localhost does not have access to the isLoggedIn cookie set after successful login, there's a piece on the /apply/logged-in container that sets this cookie to true if in development mode


server/controllers/auth.js
```diff
@@ -5,13 +5,19 @@ const authNew = (passport) => {
...
+  return passport.authenticate('oauth2', { failureRedirect: '/auth/error'});
+};
 
 const authSuccess = (req, res) => {
   req.session.user = req.user; // is this right? or should be use a method in passport to do the serialization?
   res.cookie('isLoggedIn', true);
-  res.redirect('/apply/id-and-license/my-basics/legal-name');
+
+  if (process.env.APP_ENV === 'development') {
+    res.redirect('http://localhost:3000/apply/logged-in');
```
kyliepace: this is so that the loggedIn page can have access to the appName cookie

client/containers/get-started/logged-in-page.jsx
```diff
@@ -0,0 +1,29 @@
...
+  buildCookie,
+  getAppNameCookie
+} from '../../helpers/data/cookies';
+
+const loggedIn = (props) => {
+
+  // set isLoggedIn key for localhost
+  if (APP_ENV === 'development') {
+    document.cookie = 'isLoggedIn=true;path=/';
+  }
```
kyliepace: since when in dev mode the server sets this cookie to the lvh.me domain, I added this so that the localhost will still mimic the experience of being logged in

baccigalupi:
Hey @kyliepace this has been open for a while and I have looked at it a couple times. Maybe tomorrow in standup you can grab someone for a slack call to go over it. For me it is too hard to understand in this form. I would need an explanation. Guessing that may be why it hasn't been approved yet.

PR #535, opened 2018-03-22
Adds translations to the language radio buttons

baccigalupi:
The original design had Native versions of each language and the per language translations underneath. This commit does that so that Mak and co can tinker with better ways of designing the system.


kyliepace:
I fixed the error with the other selector and tested it out with the extra <div className='row'> and props.children. the change does affect the child text input margins, but should be fixable with css.

PR #536, opened 2018-03-22
log user out after 2 minutes of inactivity

kyliepace:
- added auto-logout handler (based on [stack overflow answer](https://stackoverflow.com/questions/38143993/session-auto-logout-after-inactivity))

- handler gets called on apply/logged-in page that will be called by the server once the CDL sign-in story is through

- auto-logout sets a 2 minute timeout gets reset after every action in the events array.
`this.events = ['load', 'mousemove', 'mousedown', 'click', 'scroll', 'keypress'];`


client/helpers/handlers/auto-logout.js
```diff
@@ -0,0 +1,64 @@
+// from demux's answer at https://stackoverflow.com/questions/38143993/session-auto-logout-after-inactivity
+
```
baccigalupi: This is a cool class. Listening on `mousemove` is really processor expensive. It will likely lead to some performance issues. What I was really looking for was just that the data hadn't changed in 2 minutes. But other kinds of activity might be good to catch too.

I would take `mousemove` out, and maybe look for events related to touch, since that is a thing.

client/presentations/page.jsx
```diff
@@ -39,7 +38,7 @@ const ApplicationHeader = (props) => {
 }
 
 const Logout = (props) => {
-  let isLoggedIn = cookie.load('isLoggedIn');
+  let isLoggedIn = document.cookie.replace(/(?:(?:^|.*;\s*)isLoggedIn\s*\=\s*([^;]*).*$)|^.*$/, "$1");
```
baccigalupi: Oh, so this is the cookie fix. Might be good to extract this cookie replacement into its own module. That way when we need to with language choice, we should be able to make `isLoggedIn` a configurable key to search and replace.

client/helpers/handlers/auto-logout.js
kyliepace: good idea! will do

PR #540, opened 2018-03-23
be prompted to sign in when I have accidentally left the reservation

kyliepace:
- previous PR that logs user out after 2 minutes of inactivity redirects the url to the sign-in page, so my thought is that the only way to get to the protected pages after being logged out should be by either hitting the server with a url from a bookmark or from typing in the url, or from hitting the back button

- the server checks to see if url being retrieved is protected, and if the user is logged in, or if the app_env is development

- the navigateOnBack function performs a similar check

- I have mocked the sign-in flow in acceptance tests by adding app_env === 'test' to the hack on /containers/logged-in-page. This is to set the logged-in cookie to true just by visiting this page (whereas in production the cookie is set by the server after passport is successful). On the acceptance tests where the user should be logged in, node horseman first visits the logged-in page to set the cookie to true.

- the cdl sign-in story that redirects to the correct legal name page is not yet implemented so there is some commented-out code that won't work until that story is implemented


client/containers/cdl/routes.jsx
```diff
@@ -35,15 +31,8 @@ const CDLRoutes = () => {
...
-      <Route path={ cdlPath('/disclaimers') }                          component={Disclaimers} />
-
       <Route path={ cdlPath('/sign-in') }                               component={IDMe} />
-
-      <Route path={ cdlPath('/true-name')}                              component={LegalName}  />
-      <Route path={ editCDLPath('/true-name')}                          component={LegalName}  />
-
-      <Route path={ cdlPath('/date-of-birth')}                          component={DateOfBirth}  />
-      <Route path={ editCDLPath('/date-of-birth')}                      component={DateOfBirth}  />
+      <Route path={ cdlPath('/disclaimers') }                           component={Disclaimers} />
```
kyliepace: @teresavshen should this disclaimers still be here or are we using just the one in the main routes file?

PR #545, opened 2018-03-26
Translation refactoring - wraps translational text within Translator component.

logicgraphy:
- There is a common theme for wrapping translational text within <Translator /> component.
- If the translation for given key is not found, it will return the key itself instead of a blank string. This is required for time being because of common pattern implemented across shared components. For example, if translation Key is not available for a particular text, you can pass the text to Translator component and it will show up as is. 
- Moved the text translation part out of helper functions (validations). Validators will return the error keys and those will be translated inside the presentational components. Fixed corresponding unit tests.
- Removes locale specific translations from all the pages. 
- As a side effect, emoji debugging is gone. It will be reintroduced in a following PR after this code is merged in.


client/presentations/back-button.jsx
```diff
@@ -1,18 +1,17 @@
...
       <button
-        className='arrow-button backwards'
-        onClick={props.onBack}
-        type='button'
+        className = 'arrow-button backwards'
+        onClick   = { props.onBack }
+        type      = 'button'
       >
-        {translations[locale].shared.navigation.back}
+        <Translator tag = 'span' translationPath = 'shared.navigation.back' />
```
baccigalupi: We could use a div.unit here which serves the same purpose of not taking a full line. Spans aren't being used in our css universe, to reduce complexity.

client/presentations/cdl/license-class/class-radios.jsx
```diff
@@ -16,22 +14,18 @@ const ClassRadios = (props) => {
...
+          <RadioSelector value = 'classA'>
+            <Translator tag = 'span' translationPath = 'cdl.licenseClassPage.values.0' />
+          </RadioSelector>
+
+          <RadioSelector value = 'classB'>
+            <Translator tag = 'span' translationPath = 'cdl.licenseClassPage.values.1' />
+          </RadioSelector>
+
+          <RadioSelector value = 'classC'>
+          <Translator tag = 'span' translationPath = 'cdl.licenseClassPage.values.2' />
```
baccigalupi: indentation?

PR #547, opened 2018-03-27
edit to auth protection strategy

kyliepace:
- clear language cookie from choose-language container
- auto-logout makes fetch to '/apply/log-out' route on server
- previous strategy to keep logged-out users out of protected pages was not sufficient, so added a protected-route component to most routes that redirects to sign-in page if user is not logged in and in production env
- moved the id-and-license routes into own folder
- note that dev, test, and acceptance app_envs will not be affected and will let logged-out users see all the pages


client/helpers/handlers/auto-logout.js
```diff
@@ -41,8 +42,20 @@ class AutoLogout {
   }
 
   logout() {
-    this.history.push(`/apply/${this.appName}/log-out`);
-    this.destroy();  // Cleanup
+    let that = this;
+
+    let redirect = `/apply/${this.appName}/sign-in`;
```
kyliepace: include a fetch to logout route

client/containers/private-route.jsx
```diff
@@ -0,0 +1,23 @@
...
+  let appName = getAppNameCookie();
+  let pathNameURL = signInURL(appName);
+
+  return (
+    <Route
+      path={pathURL}
+      render={(props) => requireLogIn(props)
+        ? <Redirect to={{pathname: `${pathNameURL}`}} />
+        : <Component {...props} />
+      }
```
kyliepace: PrivateRoute checks to see if user should be logged in to see page and if they are not logged in, it redirects to the sign-in page

client/helpers/handlers/navigate-on-back.js
```diff
@@ -11,21 +9,13 @@ const normalFlowOrValid = (props, validator) => {
   }
 };
 
-const onBackGenerator = (props, validator, env=APP_ENV, cookie=document.cookie) => {
+const onBackGenerator = (props, validator) => {
```
kyliepace: removed the signIn function from onBackGenerator since routes are protected by PrivateRoute instead

client/helpers/handlers/auto-logout.js
kyliepace: not sure how to test this though

PR #557, opened 2018-03-28
removed server require-log-in controller

kyliepace:
- staging site was being redirects to /apply/logged-in/sign-in after successful login

- first commit I edited the server require-log-in controller since it seems like that was the source of the bad redirect

- then I decided that since routes are already protected on client side, server-side is redundant

- moved the client-side require-log-in helpers from data/cookies to data/application


client/helpers/data/application.js
```diff
@@ -19,4 +20,24 @@ export const ballotLanguageIsSelected = (ballotLanguage) => {
...
+    }
+  }
+  return true;
+};
+
+export const isProduction = (env) => {
+  return  env !== 'development' &&
+          env !== 'test' &&
+          env !== 'acceptance';
+};
```
logicgraphy: I think this might potentially cause issue because there might be application environments that are named different from here and still not a production environment. For example, 'staging' or 'qa'. 
In the server side configuration, env variables are loaded from .env file if the env is 'development' or 'test'. So probably we can keep this consistent with how server is configured. That way, we can bypass required-login checks in these two environments for convenience purposes.

PR #558, opened 2018-03-29
Enables emoji debugging on page component

logicgraphy:


client/helpers/handlers/on-emoji-debug.js
```diff
@@ -0,0 +1,13 @@
...
+
+import getTranslation     from '../../actions/get-translation';
+import { updateLanguage } from '../../actions';
+
+export default (dispatch) => {
+  return (value) => {
+    dispatch(updateLanguage('language', value));
+    if( value === 'emoji') {
+      getTranslation(value)(dispatch);
+    }
```
kyliepace: ah this is good so the translation will load instantly for emoji, as opposed to the more data-sensitive strategy we have for other languages. thumbs up.

PR #572, opened 2018-04-02
Open apps page

kyliepace:
- successful login sends user id to client apply/logged-in/:user url

- client makes fetch to api to get status of all apps associated with user ID in db (alternatively, maybe this data could be sent to user as cookie after logging in)

- if client has only one open app, page redirects to legal name page

- if client has multiple open apps, page redirects to /apply/open-apps page

- /apply/open-apps page maps all the apps and has a PageSummaryLink button that calls handlers.onFlowChange when clicked and links to the legal name page for the selected app.

- modified postData dispatch to include the user ID in the req.body

- things got a little bit boggy when dealing with how to use PageSummaryLink to update state so that user is on correct app flow experience if getting both new ID and new DL. the standard behavior is to pass the first value in the cardType array to the redux state. However, there is one situation in which the flow can include both cards at the same time, when the user is getting both new ID and DL. 
I added logic to pass ['ID', 'DL'] to the array if both of these cards are new. 
The state.application.cardType and state.application.cardAction states should probably be moved to the ui section.


kyliepace:
this is passing locally and failing on travis

PR #573, opened 2018-04-02
Server cookies

kyliepace:
save appName cookie from server instead of client, see if this fixes issue on edl-test domain


client/containers/id-me-page.jsx
```diff
@@ -14,14 +14,19 @@ const Page = (props) => {
   let onBack = handlers.navigateOnBack(props);
 
   let appName = getAppType(props);
-  buildAppName(appName);
+
+  if (APP_ENV === 'test' || APP_ENV === 'development') {
+    buildAppName(appName);
```
kyliepace: still needed for testing

client/presentations/get-started/id-me-page.jsx
```diff
@@ -18,7 +18,7 @@ const Presentation = (props) => {
         </h2>
 
         <fieldset role='group' className='id-me-buttons' aria-label='Authentication buttons'>
-          <a href='/auth/new' className='button green translation-missing id-me-create'>
+          <a href={`/auth/new/${props.appName}`} className='button green translation-missing id-me-create'>
```
kyliepace: pass appName to server so server sets cookie instead. see if this works.

server/controllers/logout.js
```diff
@@ -1,10 +1,13 @@
...
 
 const logout = (req, res) => {
+  console.log('log out');
+  let appName = req.cookies.appName;
   req.logout();
   req.session.destroy();
   res.cookie('isLoggedIn', false);
-  return res.send();
+
+  res.redirect(`/apply/${appName}/sign-in`);
```
kyliepace: testing on heroku-qa indicates that res.redirect also sends the changed cookies. confirm this works on other domain.

PR #575, opened 2018-04-03
Server cookies

kyliepace:
- realized that we can pass the appName and language data through the passport authorization via a state string and access it in the callback through req.query.state!

- I temporarily changed the APP_ENV of [dlite-web-dev branch](https://dlite-web-dev.herokuapp.com) to production and was working there (have since changed back)


server/routes.js
```diff
@@ -9,15 +9,17 @@ const routes = (passport) => {
   router.post('/api/application',         controllers.postApplication);
   router.get( '/api/translation/:code',   controllers.getTranslation);
 
-  router.get( '/auth/new',                controllers.authNew(passport));
+  router.get( '/auth/new/:appName/:language',    controllers.authNew(passport));
```
logicgraphy: Good stuff!

server/controllers/auth.js
```diff
@@ -1,25 +1,29 @@
 'use strict';
 
 const authNew = (passport) => {
-  return passport.authenticate('oauth2', { scope: ['multifactor'] });
+  return (req, res, next) => {
+    let dataString = JSON.stringify({appName: req.params.appName, language: req.params.language});
+    return passport.authenticate('oauth2', { scope: ['multifactor'], state: dataString })(req, res, next);
```
logicgraphy: Nice!

PR #578, opened 2018-04-03
Open apps

kyliepace:
- made [csrf cookies](https://github.com/expressjs/csurf) all httpOnly but would like to think more about security
- server sends user id to client. logged-in page gets summary of all open apps for that user id
- if user has multiple applications open, redirects to open-applications page
- shows name, summary of application, and button to edit application
- button is the PageSummaryLink used in the conclusion summary page
- server parses cardType for IDDL apps. if user is getting both new cards, then the server returns ["ID", "DL"], otherwise if the user is getting both cards the cardType starts with ID and the user can edit the DL from the summary page
- if error retrieving application from server, keep user on open-applications page and show error message
- need help writing testing for the onFlowChange handler


.gitignore
```diff
@@ -12,3 +12,4 @@ server/config/database.json
 public/index.html
 public/index.dev.html
 public/index.stage.html
+public/app.js.gz
```
logicgraphy: This file should live outside of repo

PR #586, opened 2018-04-03
added option so that if no cookieLanguage is saved, the default engli…

kyliepace:
…sh gets saved to redux after auth flow


client/helpers/handlers/load-translation-from-cookie.js
```diff
@@ -12,7 +12,8 @@ const loadTranslationFromCookie = (dispatch) => {
     if (!languageIsSelected(language)) {
       let cookieLanguage = getLanguageFromCookie();
       // save language from cookie to redux state.ui.language
-      dispatch(updateLanguage('language', cookieLanguage));
+      let saveLanguage = cookieLanguage || 'en';
+      dispatch(updateLanguage('language', saveLanguage));
       if (!doNotNeedToLoadTranslations(cookieLanguage)) {
```
kyliepace: the doNotNeedToLoadTranslations helper will keep the translation from loading if language is blank or is 'en'

PR #596, opened 2018-04-04
Bug: Suffix dropdown for Internet Explorer

teresavshen:
- https://www.pivotaltracker.com/n/projects/2051881/stories/156448379


kyliepace:
looks like a unit test is failing, is the only thing keeping this back

PR #623, opened 2018-04-06
explicitly save session

kyliepace:
[based on this github discussion](https://github.com/jaredhanson/passport/issues/482)
- redirect on logout only as callback to session being destroyed


LeDeep:
code climate score seems odd. wonder why it's so low for the diff-coverage

PR #667, opened 2018-04-13
added TST_SPLASH_SCREEN_URL env variable

kyliepace:


kyliepace:
this is failing on the acceptance tests at the same places the current no-auth branch is also failing

PR #674, opened 2018-04-14
Add feature that saves to database as user goes through app. 

LeDeep:
- Saves application while user is completing it
- The problem we were having before was using `setInterval`. `setInterval` would be set every time the component was mounted, so it ran more often than we wanted.
- This was causing multiple records to be saved for each user
- Initial save happens after user comes to address page, and every section from then on


client/containers/auto-save.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+  componentDidMount() {
+    let app = this.app;
+    let state = app.state;
+    let handlers = this.handlers;
+    if(app.chooseApp === 'iddl' && state.application.cardAction) {
+      handlers(state.application)
+    } else if (app.chooseApp === 'cdl' && state.cdl.cardAction) {
+      handlers(state.cdl)
+    };
```
kyliepace: also pass along userID and pathname to the postApplication action

client/reducers/api-reducer.js
```diff
@@ -7,10 +7,18 @@ const apiReducer = (state, action) => {
...
-      let applicationState            = Object.assign({}, state);
-      applicationState.application.id = applicationID;
-      return applicationState;
+      if(state.ui.chooseApp === 'iddl') {
+        let applicationID               = action.payload.data.application_id;
+        let applicationState            = Object.assign({}, state);
+        applicationState.application.id = applicationID;
+        return applicationState;
+      }
+      if(state.ui.chooseApp === 'cdl') {
```
kyliepace: we might also have to change the appointment prep helper that generates the confirmation code to look at the right place

client/containers/auto-save.jsx
```diff
@@ -0,0 +1,45 @@
...
+    super(app);
+    this.app = app;
+    this.handlers = handlers.autoSave(this.app.dispatch);
+  }
+
+  componentDidMount() {
+    let app = this.app;
+    let state = app.state;
+    let handlers = this.handlers;
+    if(app.section.key !== getTextFromState(app, app.sectionKey, '')) {
```
kyliepace: nice! might need to add more logic so that auto-save doesn't happen when user goes from logged-in to the open-applications page, and when they go from that page to an existing application. could check that app.section.key and the getTextFromState result are not undefined, and that it is not an intro <---> summary change?

PR #683, opened 2018-04-15
Timeout no auth

kyliepace:


client/actions/log-out.js
```diff
@@ -1,28 +1,32 @@
...
   return (dispatch) => {
     dispatch({
       type: 'UPDATE_LOGGED_IN',
       payload: {
         value: false
       }
     });
-    return fetch('/apply/log-out', {
+
+    if (TST_ENV.toString() === 'true' && tstData.adaTst.toString() === 'false') {
```
kyliepace: if a TST terminal that's not an ADA, push the url to location now. otherwise cors fails and url is not opened on auto-logout

client/actions/log-out.js
```diff
@@ -1,28 +1,32 @@
...
+      location.href = tstData.splashURL;
+    }
+
+    return fetcher('/apply/log-out', {
       method: 'GET',
       credentials: 'same-origin',
       headers: {
         'Accept': 'application/json',
-        'Content-Type': 'application/json'
+        'Content-Type': 'text/plain'
```
kyliepace: did this to avoid preflight

PR #698, opened 2018-04-22
Buttons

teresavshen:
- Removed SVG buttons due to load issues
- Replaced with CSS boxes


kyliepace:
let's be sure to get this into no-auth as well

PR #717, opened 2018-04-25
do not autosave when on open-applications page 

kyliepace:
application was trying to auto-save when user goes from id.me to the /open-applications page, and again when the user selects an application to resume it. This is not ideal in the latter case and makes no sense in the former, but shows that the auto-save feature is doing a good job saving on changing sections! too good of a job!

this won't affect the TST experience but would affect the public app


server/controllers/post-application.js
```diff
@@ -10,6 +10,8 @@ module.exports = function postApplication(req, res) {
     data.id = uuidv4();
   }
 
+  console.log('saved application ' + data.id);
```
logicgraphy: probably not a good idea to print identifiers in the log file.

PR #740, opened 2018-05-03
adjusted voting parser so that decline to answer is saved 

kyliepace:
- "decline" answer will be saved to db and accessible to phoebe
- "politicalPartyChoose" server-to-client parser will show blank string if user did not get to that question

currently if the user declines to answer citizenship or eligibility question, no row is saved to the voting table and phoebe can not discern whether user has selected "decline" or "No" to questions


LeDeep:
Looks good. Can be merged once conflicts are cleared.

PR #753, opened 2018-05-09
Resolve issue where card id changes on every save, which was causing …

LeDeep:
… multiple card options records and multiple license class records
- The `ids` in the `cards` table were incrementing with every save, which caused multiple records for other tables, namely 'card_options' and 'license_classes'
- This fixes that issue, so `ids` in the `cards` no longer increment and multiple records are not created


server/models/db/post-application.js
```diff
@@ -53,14 +53,13 @@ function insertCards(cards, application_id) {
...
-              .insert(cards)
-              .returning("*");
-          });
+        let cardData = cards.map(card => {
+          return db("cards")
+            .where("application_id", application_id)
+            .update(card)
+            .returning("*");
+        });
+        return Promise.all(cardData);
```
kyliepace: cool! I think we wrote the previous code with the concern that the number of cards saved in the db could vary from update to update, but maybe having the index keeps the update limited to each combination of application_id and card id. the tests we wrote around those cases are passing but we could always have more

PR #754, opened 2018-05-09
Feature toggle integration

kyliepace:
- toggles features according to APP_MODE ('public', 'tst', or 'ada') variable set on build time
- run-time variables like timeout are phased out in favor of build time
- fix for saving multiple cards in db broke acceptance test in this branch so I reverted back to the previous method, let's look more into how to implement a cleaner solution and add more testing around this


client/helpers/data/application.js
```diff
@@ -51,8 +51,8 @@ export function requireLogIn(pathname, isLoggedIn, env = APP_ENV) {
   return isProduction(env) && !isLoggedIn && afterIntro(pathname);
 }
 
-export const hasMultipleApps = props => {
-  return props.userData.appsLength > 0;
+export const hasMultipleApps = (props, env = APP_MODE) => {
```
logicgraphy: Only for the sake of consistency, we should stick with referencing APP_ENV as env. 
So, `mode = APP_MODE` maybe?

kyliepace: yep! will change

PR #755, opened 2018-05-09
extract cards insert/update function to separate file and add testing

kyliepace:
Deep's change this morning found the problem and fixed it, but when I ran the auth-flow acceptance test only one of the card types was being saved and it was being saved in both rows in the cards table.

Here I've spelled out all possible combinations (that I thought of - please add if more) and how they should be handled, and written testing. It could be overkill for production but could help database query refactoring?


LeDeep:
This seems like overkill. I think we just need to do a `where` on the card type. I'll put up a pr and include the additional tests you added.

LeDeep:
Here's the pull request: https://github.com/stateofca/dmv-alice/pull/758

I added the tests you wrote and those are passing. Let me know if the pr resolves any of the other issues you were seeing.

PR #757, opened 2018-05-10
Fixes issue where composeWithDevTools was undefined

LeDeep:


client/containers/app.jsx
```diff
@@ -9,6 +9,7 @@ import createBrowserHistory from "history/createBrowserHistory";
 import thunk from "redux-thunk";
 import reducers from "../reducers";
 import Routes from "./routes.jsx";
+import { composeWithDevTools } from "redux-devtools-extension/developmentOnly";
```
logicgraphy: Dev dependencies will not be installed for production, so this will need to be imported only when in development env.

PR #761, opened 2018-05-10
Resolve issue where logged-in page was returning an error on tst env

LeDeep:


logicgraphy:
Will be nice to have some unit tests around this to make sure the dispatch() is working as expected in public, tst and ada alice.

PR #765, opened 2018-05-11
finalized table

kyliepace:
new migration to add a table called finalized that has id, application_id, and finalized_at timestamp. saves with application and is pulled in user data query


client/reducers/server/user-data.js
```diff
@@ -10,7 +10,8 @@ function defaultState() {
         cardType: "",
         name: "",
         id: "",
-        updatedAt: ""
+        updatedAt: "",
+        finalizedAt: null
```
logicgraphy: finalizedAt should be associated at the application level, as one user can have multiple applications and not all will be finalized at ones. Not sure what is the intent of this in user data reducer

server/db/migrations/20180511133029_finalized_at.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+exports.up = function(db, Promise) {
+  return db.schema.createTable('finalized', table => {
+    table.increments('id').primary();
+    table.string('application_id').index();
+    table.date('finalized_at');
```
logicgraphy: finalized_at Should be a timestamp

logicgraphy: finalized_at Should be a timestamp

server/models/db/post-application.js
```diff
@@ -395,6 +395,12 @@ function saveApplication(data) {
             if (records.length > 0) {
               returnedData.guardian_addresses = records;
             }
+          }),
+        insertOneToOne('finalized', {application_id: application_id, finalized_at: null}, application_id)
```
logicgraphy: finalized table should never be populated from Alice application, not sure why would we need this here.

server/db/migrations/20180511133029_finalized_at.js
kyliepace: how about dateTime?

client/reducers/server/user-data.js
kyliepace: agreed. here it's added to each application associated with the user. it needs to get pulled here so that the open-applications page can access the finalizedAt value

server/models/db/post-application.js
kyliepace: removed

client/reducers/server/user-data.js
logicgraphy: Got it!

server/db/migrations/20180511133029_finalized_at.js
LeDeep: dateTime would be consistent with what we use for `created_at` and `updated_at`

server/models/db/post-application.js
kyliepace: actually I think it needs to create a row in this table so that when herb does a join it doesn't return empty.

logicgraphy: Right.
Alice will do inserts for each application with application_id.
Herb will do an update for finalized_at

PR #766, opened 2018-05-11
open-applications page filter out finalized apps

kyliepace:
applications where finalizedAt is not null are not shown on open-applications page


logicgraphy:
It looks good. We need to work with product and find out the appropriate message/action to show on UI when an application is finalized.

